PR #1, opened 2020-03-06
Tide example

dusty-phillips:
Just stumbled across async-graphql. Looks like you've done a lot of great work in a short amount of time. Nice work. 

I wanted to experiment with using async-graphql with tide in my project. It seems to work fine, as the attached example shows.


sunli829:
Thank you for your contribution!

vkill:
[here](https://github.com/sunli829/async-graphql-examples/pull/1) is the latest example.

PR #7, opened 2020-03-20
Add support for bson crate's ObjectId as a Scalar, so that GraphQL in…

iancormac84:
…tegration with MongoDB becomes possible.


sunli829:
thx😁

PR #10, opened 2020-03-29
ExtendedError and ResultExt inital draft

### LINKED ISSUE #9: Support Error Extensions
Hey,
first of all, I really enjoy using your library so far. I think it is already very comprehensive. One thing I am missing though is returning extensions for errors as described [here](https://github.com/graphql/graphql-spec/blob/master/spec/Section%207%20--%20Response.md) in the spec. Something like: 
```{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ],
      "extensions": {
        "code": "CAN_NOT_FETCH_BY_ID",
        "timestamp": "Fri Feb 9 14:33:09 UTC 2018"
      }
    }
  ]
}
```

In terms of api, I think it would be nice to have a trait with a single method `with_extensions` which you could implement on your errors. Something like: 
```
// somewhere in async-graphql
pub trait ErrorExtensions {
    fn with_extensions(&self) -> Option<serde_json::Value> {
        None
    }
}

impl<T: Into<Error>> ErrorExtensions for T {}
```
```
// somewhere in users code...
impl ErrorExtensions for MyError {
  fn with_extensions(&self) -> Option<serde_json::Value> {

    let extensions = serde_json::json!({
       "code": self.code.to_string(),
       "timestamp": get_current_timestamp()
    });

    Some(extensions)
  }
}
```

I am happy to contribute, so if you want any help just let me know.

nicolaiunrein:
My first draft on implementing extensions to errors which should resolve #9. Example needs some cleanup and documentation. Tests are also not complete. This is more about discussing the general idea and the provided interface for the users. I had to drop the original idea of a simple trait because without specialization-feature [RFC 1210](https://github.com/rust-lang/rust/issues/31844#) it would get really hacky to implement user overwriteable methods.

Instead I used the wrapper approach like `PositionError` plus an additional convenience trait on the `std::Result`. 

Have you considered replacing the re-exported anyhow::Error type with some custom error you are in control of? I think this could simplify the overall error-handling code a little bit.

### DIFF
--- a/examples/error_extensions.rs
+++ b/examples/error_extensions.rs
@@ -1,0 +1,108 @@
+use actix_rt;
+use actix_web::{guard, web, App, HttpResponse, HttpServer};
+use async_graphql::http::{graphiql_source, playground_source, GQLRequest, GQLResponse};
+use async_graphql::*;
+use serde_json::json;
+
+#[derive(Debug)]
+pub enum MyError {
+    NotFound,
+    ServerError(String),
+    ErrorWithoutExtensions,
+}
+
+// Let's implement a mapping from our MyError to async_graphql::Error (anyhow::Error).
+// But instead of mapping to async_graphql::Error directly we map to async_graphql::ExtendedError,
+// which gives us the opportunity to provide custom extensions to our errors.
+// Note: Values which can't get serialized to JSON-Objects simply get ignored.
+impl From<MyError> for Error {
+    fn from(my_error: MyError) -> Error {
+        match my_error {
+            MyError::NotFound => {
+                let msg = "Could not find ressource".to_owned();
+                let extensions = json!({"code": "NOT_FOUND"});
+                ExtendedError(msg, extensions).into()
+            }
+            MyError::ServerError(reason) => {
+                ExtendedError("ServerError".to_owned(), json!({ "reason": reason })).into()
+            }
+
+            MyError::ErrorWithoutExtensions => ExtendedError(
+                "No Extensions".to_owned(),
+                json!("This will be ignored since it does not represent an object."),
+            )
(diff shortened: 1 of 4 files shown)


nicolaiunrein:
Resolves #9

sunli829:
Thank you for PR.😁
This should be the best solution for now.

Replace anyhow::Error with custom error, may be less important, because `async-graphql` is the server side, just export json.
Do you have any suggestions for this?

nicolaiunrein:
You're welcome!
I am very excited to implement the new feature into my microservice on monday. 😁 I just recently started using juniper when I discovered your crate on friday and I instantly preferred it over juniper. Considering the very young age of your repo this is really amazing. I think your work has great potential becoming the de facto standard for rust GQL code. Keep it up!

If there is anything else I can do to support just let me know. I think adding documentation is really important to make the crate more user-friendly/popular but I am not a native speaker so I don't know if I can help with that at all.

Regarding the Error type:
If I'm correct, the way it is implemented leads to a deeply nested struct:
original Error -> anyhow::Error -> ExtendedError -> anyhow::Error -> PositionError -> anyhow::Error.

Just thought it might be easier and more maintainable (internally) to have a wrapper struct with optional private fields for position and extension (and what might come in the future) but didn't know if it is applicable or at all sensible. Just something to consider.

From the users point of view this should not make much of a difference.

sunli829:
Well, the main problem now is to improve the documentation, I just don't have enough time to do it now.
If you have any suggestions for improvement, please let me know.

PR #12, opened 2020-03-30
Add documentation to Upload type

nicolaiunrein:
Added some documentation to the upload type taken from the example. I also included an example curl command for testing.


sunli829:
Thank you!😁

PR #23, opened 2020-04-22
Add failing interface test

phated:
I'm adding a failing test for Interfaces resolvers being applied when an Object doesn't belong to it.

I just ran into this in my project and I believe this is the smaller version.

### DIFF
--- a/tests/interface.rs
+++ b/tests/interface.rs
@@ -161,3 +161,83 @@ pub async fn test_multiple_interfaces() {
         })
     );
 }
+
+#[async_std::test]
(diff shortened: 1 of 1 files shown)


sunli829:
I'm looking at the problem.

sunli829:
This bug is hard to find. Thanks for the test code.😁
Please upgrade to `v1.9.10`，I solved the problem.

phated:
Thank you so much! It seems to be working now. I appreciate the speed you are able to fix these!

PR #26, opened 2020-04-23
Add real-world example

phated:
@sunli829 again, thank you so much for being very responsive to my issues and requests! 🙇

I've finally pushed my changes that replace Juniper with async-graphql and I wanted to highlight a real-world example of using your library! If you don't think this should be in the README, feel free to close this PR.

Keep up the great work!


sunli829:
Thank you for your contribution to 'async-graphql'.😁

PR #27, opened 2020-04-24
Add bson::UtcDateTime support

IcanDivideBy0:
Simple support for [bson::UtcDateTime](https://docs.rs/bson/0.14.1/bson/struct.UtcDateTime.html) as scalar.


sunli829:
Thank you very much for you PR.

PR #46, opened 2020-05-03
Implement OutputValueType for FieldResult

IcanDivideBy0:
Hello,
While using async_graphql I've came across several usecase where it's useful to return a FieldResult as an output type:

1. ` -> Option<FieldResult<_>>`
1. ` -> Vec<FieldResult<_>>`
1. ` -> impl Stream<Item = FieldResult<_>>`

While the first two can be easily handled by swaping things around, the third one is a bit more tricky.

In the actual implementation, when a field fail, only the error is returned even if other fields resloved successfuly, so the first two usecase don't have much sense in this context.

When using subscriptions tho, many thing can occurs during the stream (database connection being lost for example) and right now I can't figure a way to send those errors to the client.

This PR is not really intended to be merged as is, I'd like to first discuss those usecases with you folks and get some feedback on what you think about it. Also while I tested this implementation, I've noticed the error path is lost in the response, so there's definilty more to do here.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,5 +1,7 @@
+use crate::{registry, ContextSelectionSet, OutputValueType, Type};
 use graphql_parser::query::{ParseError, Value};
 use graphql_parser::Pos;
+use std::borrow::Cow;
 use std::fmt::Debug;
 
 /// FieldError type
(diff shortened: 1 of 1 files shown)


sunli829:
I'm actually adding this feature, and I've already done some of it.😁
I thought FieldResult's implementation of OutputValueType was a good idea, but I ran into some problems with code generation at the time, and now I'll try the possibility again.

sunli829:
I made some initial attempts and this should be done quickly. 😁

sunli829:
I've completed this feature  in `1.10.4` version, and you mentioned that the third problem was solved in the previous release, so I only need to implement `Type` and `OutputType` for FieldResult to solve all the problems.

Thank you for your suggestions.😁

IcanDivideBy0:
TBH I'm really amazed by how fast you always respond on this project an the tremendous amount and quality of work you already produced!

PR #48, opened 2020-05-05
Add Cursor scalar

phated:
The [Relay spec](https://relay.dev/graphql/connections.htm#sec-Cursor) says that a Cursor can be "a custom scalar that serializes as a String, or a Non‐Null wrapper around a custom scalar that serializes as a String."

By using a custom scalar, we allow users to implement `Into<Cursor>` for their own types and it won't collide with `Into<String>` that might be implemented for another purpose.


sunli829:
Thank you.😄

PR #49, opened 2020-05-05
Export the new `Cursor` type

IcanDivideBy0:


phated:
Thanks for catching that!

PR #50, opened 2020-05-05
DataSource pagination

phated:
I was using this in my project to handle pagination, which was trying to follow the Relay spec for [pagination algorithm](https://relay.dev/graphql/connections.htm#sec-Pagination-algorithm) and I though it would be helpful to support in `DataSource`.

This adds a ton of complexity, so I'm happy to iterate on this if you want changes. I also believe the documentation could be improved, but I didn't want to commit a ton of effort if we need to change this.

Relies on #49 because I forgot to export the Cursor type.

### DIFF
--- a/src/types/connection/mod.rs
+++ b/src/types/connection/mod.rs
@@ -10,28 +10,97 @@ pub use connection_type::Connection;
 pub use cursor::Cursor;
 
 /// Connection query operation
-pub enum QueryOperation<'a> {
-    /// Forward query
-    Forward {
+pub enum QueryOperation {
+    /// Return all results
+    None,
+    /// Return all results after the cursor
+    After {
         /// After this cursor
-        after: Option<&'a str>,
-
-        /// How many records did this query return
-        limit: usize,
+        after: Cursor,
     },
-    /// Backward query
-    Backward {
+    /// Return all results before the cursor
+    Before {
         /// Before this cursor
-        before: Option<&'a str>,
-
-        /// How many records did this query return
+        before: Cursor,
+    },
+    /// Return all results between the cursors
+    Between {
+        /// After this cursor
+        after: Cursor,
+        /// But before this cursor
+        before: Cursor,
+    },
+    /// Return the amount of results specified by `limit`, starting from the beginning
+    First {
+        /// The maximum amount of results to return
         limit: usize,
     },
+    /// Return the amount of results specified by `limit`, starting after the cursor
+    FirstAfter {
+        /// The maximum amount of results to return
+        limit: usize,
+        /// After this cursor
+        after: Cursor,
+    },
+    /// Return the amount of results specified by `limit`, starting from the beginning but ending before the cursor
+    FirstBefore {
+        /// The maximum amount of results to return
+        limit: usize,
+        /// Before this cursor
+        before: Cursor,
+    },
+    /// Return the amount of results specified by `limit`, but between the cursors. Limit includes beginning results.
+    FirstBetween {
+        /// The maximum amount of results to return
+        limit: usize,
+        /// After this cursor
+        after: Cursor,
+        /// But before this cursor
+        before: Cursor,
+    },
+    /// Return the amount of results specified by `limit`, but before the end
+    Last {
+        /// The maximum amount of results to return
+        limit: usize,
+    },
+    /// Return the amount of results specified by `limit`, but before the end. Must not include anything before the cursor.
+    LastAfter {
+        /// The maximum amount of results to return
+        limit: usize,
+        /// After this cursor
+        after: Cursor,
+    },
+    /// Return the amount of results specified by `limit`, but before the cursor
+    LastBefore {
+        /// The maximum amount of results to return
+        limit: usize,
+        /// Before this cursor
+        before: Cursor,
+    },
+    /// Return the amount of results specified by `limit`, but between the cursors. Limit includes ending results.
+    LastBetween {
+        /// The maximum amount of results to return
+        limit: usize,
+        /// After this cursor
+        after: Cursor,
+        /// But before this cursor
+        before: Cursor,
+    },
+    /// An invalid query was made. For example: sending `first` and `last` in the same query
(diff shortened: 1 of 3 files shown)


IcanDivideBy0:
This is exactly the topic I was working on :D
I agree that some utility functions around Connection object would be nice!

phated:
I believe "slice" can be made much better and more efficient for scenarios where overfetching from the database is okay. I just wanted to get this proof of concept submitted so it's not hidden in my project.

IcanDivideBy0:
This is what I was about to go with:
```rust
use async_graphql::{Cursor, FieldError, FieldResult};
use std::convert::TryFrom;
use std::ops::Range;

#[async_graphql::InputObject]
pub struct ConnectionArgs {
    #[field]
    pub first: Option<i32>,
    #[field]
    pub after: Option<Cursor>,
    #[field]
    pub last: Option<i32>,
    #[field]
    pub before: Option<Cursor>,
}

impl ConnectionArgs {
    pub fn get_range<E, T>(&self, nodes: &[(Cursor, E, T)]) -> FieldResult<Range<usize>> {
        let first = self.first.map(usize::try_from).map(Result::ok).flatten();
        let last = self.last.map(usize::try_from).map(Result::ok).flatten();
        let len = nodes.len();

        match (self.after.as_ref(), self.before.as_ref()) {
            (Some(after), None) => {
                let start = nodes
                    .iter()
                    .position(|(cursor, _, _)| *cursor == *after)
                    .map(|i| i + 1)
                    .unwrap_or(0);
                let end = (start + first.unwrap_or(len)).min(nodes.len());

                Ok(start..end)
            }
            (None, Some(before)) => {
                let end = nodes
                    .iter()
                    .position(|(cursor, _, _)| *cursor == *before)
                    .unwrap_or(len);
                let start = end.saturating_sub(last.unwrap_or(len));

                Ok(start..end)
            }
            (None, None) => match (first, last) {
                (Some(first), None) => Ok(0..first.min(len)),
                (None, Some(last)) => Ok(len.saturating_sub(last)..len),
                (None, None) => Ok(0..len),

                (Some(_), Some(_)) => Err(FieldError::from(
                    "[ConnectionArgs] first and last are mutually exclusive.",
                )),
            },
            (Some(_), Some(_)) => Err(FieldError::from(
                "[ConnectionArgs] before and after are mutually exclusive.",
            )),
        }
    }
}
```

src/types/connection/mod.rs
```diff
@@ -10,29 +10,98 @@ pub use connection_type::Connection;
...
 #[async_graphql_derive::SimpleObject(internal)]
 pub struct EmptyEdgeFields;
 
+// Temporary struct for to store values for pattern matching
+struct Pagination {
+    after: Option<Cursor>,
+    before: Option<Cursor>,
+    first: Option<i32>,
+    last: Option<i32>,
+}
```
IcanDivideBy0, line 103: Having this implementing `InputObject` could be useful, so we can also use it directly in queries, ie:
```gql
query {
  users(page: { first: 10 }) { ... }
}
```

phated: Technically that's against the Relay spec, but we could move it outside the connection namespace.

IcanDivideBy0: In practice I find it convenient so I can separate pagination from filters in arguments, but that's a matter of taste I guess. You can ignore my previous comment, anyone can still go with this pattern if they want in their own project.

sunli829:
@phated Thanks for your PR, the support of the relay specification is a big step forward.😁

src/types/connection/mod.rs
sunli829: > This is what I was about to go with:
> 
> ```rust
> use async_graphql::{Cursor, FieldError, FieldResult};
> use std::convert::TryFrom;
> use std::ops::Range;
> 
> #[async_graphql::InputObject]
> pub struct ConnectionArgs {
>     #[field]
>     pub first: Option<i32>,
>     #[field]
>     pub after: Option<Cursor>,
>     #[field]
>     pub last: Option<i32>,
>     #[field]
>     pub before: Option<Cursor>,
> }
> 
> impl ConnectionArgs {
>     pub fn get_range<E, T>(&self, nodes: &[(Cursor, E, T)]) -> FieldResult<Range<usize>> {
>         let first = self.first.map(usize::try_from).map(Result::ok).flatten();
>         let last = self.last.map(usize::try_from).map(Result::ok).flatten();
>         let len = nodes.len();
> 
>         match (self.after.as_ref(), self.before.as_ref()) {
>             (Some(after), None) => {
>                 let start = nodes
>                     .iter()
>                     .position(|(cursor, _, _)| *cursor == *after)
>                     .map(|i| i + 1)
>                     .unwrap_or(0);
>                 let end = (start + first.unwrap_or(len)).min(nodes.len());
> 
>                 Ok(start..end)
>             }
>             (None, Some(before)) => {
>                 let end = nodes
>                     .iter()
>                     .position(|(cursor, _, _)| *cursor == *before)
>                     .unwrap_or(len);
>                 let start = end.saturating_sub(last.unwrap_or(len));
> 
>                 Ok(start..end)
>             }
>             (None, None) => match (first, last) {
>                 (Some(first), None) => Ok(0..first.min(len)),
>                 (None, Some(last)) => Ok(len.saturating_sub(last)..len),
>                 (None, None) => Ok(0..len),
> 
>                 (Some(_), Some(_)) => Err(FieldError::from(
>                     "[ConnectionArgs] first and last are mutually exclusive.",
>                 )),
>             },
>             (Some(_), Some(_)) => Err(FieldError::from(
>                 "[ConnectionArgs] before and after are mutually exclusive.",
>             )),
>         }
>     }
> }
> ```

@IcanDivideBy0 Now all the #[field] can be omitted, so you can define inputobject like this.

```rust
#[async_graphql::InputObject]
pub struct ConnectionArgs {
    pub first: Option<i32>,
    pub after: Option<Cursor>,
    pub last: Option<i32>,
    pub before: Option<Cursor>,
}
```

PR #53, opened 2020-05-07
Datasource context

phated:
I'd like to implement datasource for my structs, but my database connection pool is attached to the context. If we pass the context to `query_operation`, I should be able to implement this.

I also updated the book for Datasource based on my newer changes.

### DIFF
--- a/docs/zh-CN/src/cursor_connections.md
+++ b/docs/zh-CN/src/cursor_connections.md
@@ -22,25 +22,37 @@ impl DataSource for Integers {
     // 我们不需要扩展边的字段，所以传EmptyEdgeFields
     type EdgeFieldsObj = EmptyEdgeFields;
 
-    async fn query_operation(&self, operation: &QueryOperation<'_>) -> FieldResult<Connection<Self::Element, Self::EdgeFieldsObj>> {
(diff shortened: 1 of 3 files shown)


docs/zh-CN/src/cursor_connections.md
```diff
@@ -22,25 +22,37 @@ impl DataSource for Integers {
...
+            QueryOperation::LastBefore {before, limit} => {
                 // 结束数字，如果没有before参数，则为0
-                let end = before
-                    .and_then(|before| before.parse::<i32>().ok())
+                let end = before.parse::<i32>()
+                    .ok()
                     .unwrap_or(0);
                 (end - *limit, end)
             }
+            // TODO: Advise to handle all cases
```
phated, line 54: @sunli829 this needs to be translated to chinese.

phated:
@sunli829 should the book be deployed on Pull Request? I think that should only be Push to master.

sunli829:
Thank you.😁

sunli829:
> @sunli829 should the book be deployed on Pull Request? I think that should only be Push to master.

Let me change it. 😁

PR #55, opened 2020-05-07
Add From<ObjectId> for ID

IcanDivideBy0:
Since we're adding uuid conversion, I guess we can add ObjectId as well.


sunli829:
Thank you.😁

PR #67, opened 2020-05-10
slightly modified based on Sunli's reply.

DataCorrupted:
@sunli829 Resolved issue we discussed in #63 .


sunli829:
Thank you. 😁

Issue #54, opened 2020-05-07
Discussion: Should ID support From<Uuid>?


phated:
In my application, IDs are a custom scalar that wraps the Uuid type. Would it be valuable for the ID scalar provided here to have a `From<Uuid>` implementation?

sunli829:
Since you have this requirement, I think this feature makes sense. I'll add it now.😁

nicolaiunrein:
I use Uuids as well and I think it is quite common. Maybe implement it for `T: Into<Uuid>` so that it works well with the new type pattern?

nicolaiunrein:
In fact this concerns all ID conversions. If that is something desirable I will give it a shot. Should not be breaking existing implementations because all types implement `From` and `Into` themselves.

phated:
@nicolaiunrein I think that would be a good idea. When converting between ID and Cursor, I need to do `id.to_string().into()` which isn't ideal.

sunli829:
@nicolaiunrein How is this feature going to be implemented?😁

nicolaiunrein:
Sorry, I am pretty busy at the moment with other projects. I will look at it tomorrow or on monday.

nicolaiunrein:
Well, we can't implement generic `From` implementation due to conflicting implementations but I would suggest to add two traits:

```rust
pub trait ToGraphqlCursor {
    fn to_graphql_cursor(&self) -> Cursor;
}

impl<T> ToGraphqlCursor for T
where
    T: std::fmt::Display,
{
    fn to_graphql_cursor(&self) -> Cursor {
        Cursor(self.to_string())
    }
}
```

and 

```rust
pub trait ToGraphqlID {
    fn to_graphql_id(self) -> ID;
}

impl<T> ToGraphqlID for T
where
    T: std::fmt::Display,
{
    fn to_graphql_id(self) -> ID {
        ID(self.to_string())
    }
}
```

which when imported give you two methods `to_graphql_cursor` and `to_graphql_id` on anything that implements `std::fmt::Display` and allows you to instead of calling `id.to_string().into()` just `id.to_graphql_cursor`. Now this can be quite handy when you implement your own ID type as well:
```rust
pub struct MyID(Uuid);
impl std::fmt::Display for MyID { ... }

// somewhere else
let id = my_id.to_graphql_id();
```

What do you think?

sunli829:
I think it's good, it's really easier to use. 😁

[...]

PR #71, opened 2020-05-10
Simplify ID and Cursor conversion

IcanDivideBy0:
As discussed in #54


IcanDivideBy0:
`fmt::Display` was actually used in `Type::global_id` so I added a `as_str` method on ID to keep the same behavior.

IcanDivideBy0:
I actually removed the `as_str` since `ID` can be deref into a String, there's no need for that.

sunli829:
Thank you. 😁

PR #72, opened 2020-05-10
Prefer TryFrom to TryInto for ID

IcanDivideBy0:
#71 was merged a bit fast :D
Generally speaking `TryFrom` is more general than `TryInto` (cause `TryFrom` implies `TryInto`), so this should gives a bit more flexibility, like using the try_from as a closure
```rust
vec![ /* ... */ ]
    .iter()
    .map(ID::try_trom)
    .collect::<Result<usize, _>>()
```

### DIFF
--- a/src/scalars/id.rs
+++ b/src/scalars/id.rs
@@ -1,7 +1,7 @@
 use crate::{InputValueError, InputValueResult, Result, ScalarType, Value};
 use async_graphql_derive::Scalar;
 use bson::oid::{self, ObjectId};
-use std::convert::TryInto;
+use std::convert::TryFrom;
 use std::num::ParseIntError;
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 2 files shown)


sunli829:
All right.😂

IcanDivideBy0:
@sunli829 don't merge this right now

IcanDivideBy0:
@sunli829 that's good now.
Since `ID` does not implement `fmt::Display` anymore, converting an `ID` to a cursor required more steps, now we can also convert `ID` into a `Cursor` easily which I believe is also a common use case.

IcanDivideBy0:
@sunli829 Thanks a lot for the invitation to collaborate on the project, I really appreciate! :D

sunli829:
Let's work together on rust's best graphql library.🤝

PR #74, opened 2020-05-11
Initial add prelude module.

sunli829:
@nicolaiunrein  @phated @IcanDivideBy0 Please review. 😁

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,7 @@ mod subscription;
 mod types;
 mod validation;
 
+pub mod prelude;
 pub mod extensions;
 pub mod guard;
 pub mod validators;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,9 +94,9 @@ mod subscription;
 mod types;
 mod validation;
 
-pub mod prelude;
 pub mod extensions;
 pub mod guard;
+pub mod prelude;
 pub mod validators;
 
 use async_graphql_parser as parser;
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -1,0 +1,18 @@
+//! Re-exports most common types with an extra GQL* prefix to prevent name clashes.
+
+// types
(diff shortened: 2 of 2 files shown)


phated:
I like it.

nicolaiunrein:
The exports look about right and I think renaming is a good idea. But I think I would prefer "GQL_ID" over "GQLID" which still looks like a const but is a little better to read. Nevertheless I would stick to whatever the naming convention is.

sunli829:
Thanks, let me change it to GQL_ID.

IcanDivideBy0:
I have mixed feelings :/
I'd love to see a prelude mod but having different names in the prelude than in the crate root may confuse people?

sunli829:
I think the usual way to use prelude should be like this?

```rust
use async_graphql::prelude::*;
```

nicolaiunrein:
I agree that it might be confusing. But I think the underlying problem is that names in this library are too generic. What about prefixing all the generic pub types in this library and keep deprecated un-prefixed type-aliase around until the next major update in order not to break existing code? I think generic names like Object and Value etc should really be left for the user.

IcanDivideBy0:
I agree with @nicolaiunrein renaming those types in the root crate while still exporting them with deprecated names seems better to me.
I can prepare a PR based on this one if we all agree

sunli829:
I agree.
🍺🍺🍺🍺🍺🍺

sunli829:
@IcanDivideBy0 
I am going to complete # 15 now, the changes will be relatively large, so it is best to wait for me to complete this first, otherwise it will be painful to merge the code.

IcanDivideBy0:
@sunli829 ok!

sunli829:
@IcanDivideBy0 
It was finally done because I was stuck for a long time with a small problem, you can continue. 😁

IcanDivideBy0:
done in #78

PR #75, opened 2020-05-11
Change interfaces & unions to require enums

phated:
@sunli829 this is my first time working on a macro, so it might be bad.

I have found it weird that the `#[async_graphql::Interface]` macro replaces a `struct` definition with a `enum` definition. You can see me trying to extend an interface with diesel traits in https://github.com/phated/twentyfive-stars/blob/master/server/src/data/character_mode.rs - and you can see the "weird" behavior at https://github.com/phated/twentyfive-stars/blob/master/server/src/data/character_mode.rs#L116

By making this macro only usable on enums, we have a more natural usage pattern when extending it.

### DIFF
--- a/async-graphql-derive/src/interface.rs
+++ b/async-graphql-derive/src/interface.rs
@@ -6,26 +6,25 @@ use inflector::Inflector;
 use proc_macro::TokenStream;
 use proc_macro2::{Ident, Span};
 use quote::quote;
+use std::collections::HashSet;
 use syn::{Data, DeriveInput, Error, Fields, Result, Type};
 
 pub fn generate(interface_args: &args::Interface, input: &DeriveInput) -> Result<TokenStream> {
     let crate_name = get_crate_name(interface_args.internal);
     let ident = &input.ident;
     let generics = &input.generics;
-    let attrs = &input.attrs;
-    let vis = &input.vis;
     let s = match &input.data {
-        Data::Struct(s) => s,
-        _ => return Err(Error::new_spanned(input, "It should be a struct.")),
-    };
-    let fields = match &s.fields {
-        Fields::Unnamed(fields) => Some(&fields.unnamed),
-        Fields::Unit => None,
-        _ => return Err(Error::new_spanned(input, "All fields must be unnamed.")),
+        Data::Enum(s) => s,
+        _ => {
+            return Err(Error::new_spanned(
+                input,
+                "Interfaces can only be applied to an enum.",
+            ))
+        }
     };
     let extends = interface_args.extends;
     let mut enum_names = Vec::new();
-    let mut enum_items = Vec::new();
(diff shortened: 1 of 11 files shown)


sunli829:
Why not call `into` directly?

```rust
UpgradeMode {
          id,
          title,
          stars,
          type_,
          faction,
          traits,
          attack_modifier,
          defense_modifier,
        }.into()
```

phated:
For the creation, `.into()` would work, but I am working on a new feature where I need to pattern match on the enum variants.

If you agree to this implementation, I will update the documentation and also make changes to Union. Please let me know

phated:
This adds some extra verbosity in the definition, but I believe it makes the code easier to understand when you need to use the data types in other parts of the server.

sunli829:
Someone really asked me why `Interface` would be a struct, and I told him it was actually an enumeration. In fact, when I first implemented the `Interface`, it was an enumeration, but I was lazy. I did n’t like to write any extra code. The enumeration forced me to make a repeated name, like this `enum MyInterface {Foo (Foo)}` , so changed to struct. But it does make the code difficult to understand. Thanks for your PR, I will merge later, because I need to change the `Union` to use enumeration.

phated:
Great! I will work on `Union` now!

sunli829:
Although you have a lot of open source responsibilities, I still want to add you to the collaborators, because your suggestions are very important to this project, are you willing? 😁

phated:
> are you willing?

Yeah, I am happy to help where I can. Thank you! 🙇 I "watch" this project already, so I see all of the activity 😄

sunli829:
Great, let's do the best GraphQL library of rust together. 😄

phated:
@sunli829 the Union is implemented. I also added some tests and documentation for it. In addition, I updated the book examples.

Do you need me to update the examples or anything else?

sunli829:
Too fast, the code generation part I wrote is not very good, a little messy.😂
I will translate the book to Chinese, if you want, you can help update the Examples repo.

phated:
> Too fast, the code generation part I wrote is not very good, a little messy.😂

@sunli829 do you need me to change something? Or are you going to improve on this?

> I will translate the book to Chinese, if you want, you can help update the Examples repo.

The book doesn't mention `struct`, so I changed the example. I don't think anything else needs to be done. 

I will work on the examples repo when we finish this. Probably tomorrow because it is late for me.

Thank you!

sunli829:
> do you need me to change something? Or are you going to improve on this?
Now it seems unnecessary, after all, the amount of code is not too large. 😁

> I will work on the examples repo when we finish this. Probably tomorrow because it is late for me.
Thanks, take a break early, good night. 😁

sunli829:
I will update the examples repo after the new version is released. This version changes a lot. 😁

PR #78, opened 2020-05-11
Adding GQL prefix on important components

IcanDivideBy0:
This what I have so far
I can't find the button to mark the PR as Draft, is that something that should be enabled at project level ?

what's missing: 
- [x] `GqlError`
- [x] `GqlInputValueResult`
- [x] `GqlQueryBuilder`
- [x] `GqlResult`
- [x] `GqlValue`
- [x] `GqlID`

### DIFF
--- a/async-graphql-actix-web/src/lib.rs
+++ b/async-graphql-actix-web/src/lib.rs
@@ -7,7 +7,9 @@ mod subscription;
 use actix_web::dev::{Payload, PayloadStream};
 use actix_web::{http, web, Error, FromRequest, HttpRequest};
 use async_graphql::http::StreamBody;
-use async_graphql::{IntoQueryBuilder, IntoQueryBuilderOpts, ParseRequestError, QueryBuilder};
+use async_graphql::{
+    GqlQueryBuilder, IntoGqlQueryBuilder, IntoGqlQueryBuilderOpts, ParseRequestError,
+};
 use futures::channel::mpsc;
 use futures::{Future, SinkExt, StreamExt, TryFutureExt};
 use std::pin::Pin;
@@ -16,13 +18,13 @@ pub use subscription::WSSubscription;
 
 /// Extractor for GraphQL request
 ///
-/// It's a wrapper of `QueryBuilder`, you can use `GQLRequest::into_inner` unwrap it to `QueryBuilder`.
-/// `async_graphql::IntoQueryBuilderOpts` allows to configure extraction process.
-pub struct GQLRequest(QueryBuilder);
+/// It's a wrapper of `GqlQueryBuilder`, you can use `GQLRequest::into_inner` unwrap it to `GqlQueryBuilder`.
+/// `async_graphql::IntoGqlQueryBuilderOpts` allows to configure extraction process.
+pub struct GQLRequest(GqlQueryBuilder);
 
 impl GQLRequest {
-    /// Unwrap it to `QueryBuilder`.
-    pub fn into_inner(self) -> QueryBuilder {
+    /// Unwrap it to `GqlQueryBuilder`.
+    pub fn into_inner(self) -> GqlQueryBuilder {
         self.0
     }
 }
@@ -30,7 +32,7 @@ impl GQLRequest {
 impl FromRequest for GQLRequest {
     type Error = Error;
     type Future = Pin<Box<dyn Future<Output = Result<GQLRequest, Error>>>>;
-    type Config = IntoQueryBuilderOpts;
+    type Config = IntoGqlQueryBuilderOpts;
 
     fn from_request(req: &HttpRequest, payload: &mut Payload<PayloadStream>) -> Self::Future {
         let config = req.app_data::<Self::Config>().cloned().unwrap_or_default();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -96,6 +96,7 @@ mod validation;
 
 pub mod extensions;
 pub mod guard;
+pub mod prelude;
 pub mod validators;
 
 use async_graphql_parser as parser;
@@ -113,17 +114,17 @@ pub mod http;
 
 pub use base::{ScalarType, Type};
 pub use context::{
-    Context, ContextBase, Data, Environment, QueryPathNode, QueryPathSegment, Variables,
+    Environment, GqlContext, GqlContextBase, GqlData, GqlVariables, QueryPathNode, QueryPathSegment,
 };
 pub use error::{
-    Error, ErrorExtensions, FieldError, FieldResult, InputValueError, InputValueResult,
+    ErrorExtensions, FieldError, GqlError, GqlFieldResult, GqlInputValueResult, InputValueError,
     ParseRequestError, QueryError, ResultExt,
 };
-pub use parser::{Pos, Positioned, Value};
-pub use query::{IntoQueryBuilder, IntoQueryBuilderOpts, QueryBuilder, QueryResponse};
+pub use parser::{GqlValue, Pos, Positioned};
+pub use query::{GqlQueryBuilder, IntoGqlQueryBuilder, IntoGqlQueryBuilderOpts, QueryResponse};
 pub use registry::CacheControl;
-pub use scalars::{Any, Json, ID};
-pub use schema::Schema;
+pub use scalars::{Any, GqlID, Json};
+pub use schema::GqlSchema;
 pub use subscription::{
     SimpleBroker, SubscriptionStream, SubscriptionStreams, SubscriptionTransport,
     WebSocketTransport,
@@ -135,11 +136,11 @@ pub use types::{
 pub use validation::ValidationMode;
 
 /// Result type
-pub type Result<T> = std::result::Result<T, Error>;
+pub type GqlResult<T> = std::result::Result<T, GqlError>;
 
 // internal types
 #[doc(hidden)]
-pub use context::ContextSelectionSet;
+pub use context::GqlContextSelectionSet;
 #[doc(hidden)]
 pub mod registry;
 #[doc(hidden)]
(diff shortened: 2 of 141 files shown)


nicolaiunrein:
I wonder if it's more idiomatic to name the types eg. `GqlValue` instead of `GQLValue` like in `TcpStream` of `PartialEq` but I think because of the lowercase "L"  it gets difficult to read.

phated:
+1 to @nicolaiunrein - I like the lowercase `Gql` prefix

IcanDivideBy0:
I might have under estimated the task 😂 but it's finally done!

async-graphql-actix-web/src/lib.rs
```diff
@@ -16,21 +18,21 @@ pub use subscription::WSSubscription;
 
 /// Extractor for GraphQL request
 ///
-/// It's a wrapper of `QueryBuilder`, you can use `GQLRequest::into_inner` unwrap it to `QueryBuilder`.
-/// `async_graphql::IntoQueryBuilderOpts` allows to configure extraction process.
-pub struct GQLRequest(QueryBuilder);
+/// It's a wrapper of `GqlQueryBuilder`, you can use `GQLRequest::into_inner` unwrap it to `GqlQueryBuilder`.
+/// `async_graphql::IntoGqlQueryBuilderOpts` allows to configure extraction process.
+pub struct GQLRequest(GqlQueryBuilder);
```
phated, line 23: Should `GQLRequest` be updated too?

src/lib.rs
```diff
@@ -628,27 +635,54 @@ pub use async_graphql_derive::Union;
...
 /// Define a DataSource
-pub use async_graphql_derive::DataSource;
+pub use async_graphql_derive::GqlDataSource;
 
 /// Define a Scalar
-pub use async_graphql_derive::Scalar;
+pub use async_graphql_derive::GqlScalar;
+
+mod deprecated {
+    pub use super::GqlContext as Context;
```
phated, line 664: Is there a way to annotate deprecations in rust?

src/prelude.rs
```diff
@@ -0,0 +1,16 @@
...
+};
+
+// traits
+pub use crate::{ErrorExtensions, IntoGqlQueryBuilder, ResultExt, Type};
+
+// procedural macros
+pub use crate::{
+    GqlEnum, GqlInputObject, GqlInterface, GqlObject, GqlScalar, GqlSimpleObject, GqlSubscription,
+    GqlUnion,
+};
```
IcanDivideBy0, line 16: Here's the prelude content

src/lib.rs
```diff
@@ -628,27 +635,54 @@ pub use async_graphql_derive::Union;
...
+    pub use super::GqlSchema as Schema;
+    pub use super::GqlSimpleObject as SimpleObject;
+    pub use super::GqlSubscription as Subscription;
+    pub use super::GqlUnion as Union;
+    pub use super::GqlValue as Value;
+    pub use super::GqlVariables as Variables;
+}
+
+#[doc(hidden)]
+pub use deprecated::*;
```
IcanDivideBy0, line 688: Previous exports names are still available here, the whole project build and pass tests with this module disabled

src/lib.rs
IcanDivideBy0: I'll check that tomorrow :)

async-graphql-actix-web/src/lib.rs
IcanDivideBy0: yes I think we should change `GQLResponse` and `GQLRequest'.
There also another `GQLError` in http module, but it's just a wrapper around `&GqlError` to implement serialize, I guess we can implement `Serialize` on `GqlError` directly now

sunli829:
@IcanDivideBy0 
I was so excited when I woke up and saw this great change. It's a good idea to replace `GQL` with `Gql`, which solves the problem of `GQLID` and makes it easier to type. On behalf of async-graphql users, I thank you for your efforts.

sunli829:
The button labeled draft is indeed missing. . . There used to be. 😳

sunli829:
When I woke up and saw this huge change, I was very excited and thank @IcanDivideBy0 for his efforts.

But I also carefully considered the necessity of this Gql prefix.
Maybe it can be replaced with `use async_graphql as gql`, which should be the idiom of rust?

When they need to use prelude, I think in most cases it is `use async_graphql::prelude::*`, and then the renaming is meaningful.

So I hope we can discuss the necessity of this PR again.

I'm so sorry, this is my mistake, I did not think about it yesterday, and here I solemnly apologize to @IcanDivideBy0. 🙇

vkill:
Oh god, don't add prefix, we should use module.

phated:
Currently, `ID` in the prelude would collide with my `ID` in code because I need the newtype pattern to work with Diesel. These are the type of errors we are trying to avoid.

vkill:
@phated  Don't use `use async_graphql::*;`, you can use `use async_graphql::ID as GQLID;`

phated:
@vkill you are missing the point of a prelude. We are expected to import *

sunli829:
Renaming only the prelude export is my current thinking. 😁

crate::prelude::ID to crate::prelude::GqlID

nicolaiunrein:
I feel really sorry too 😢. So if the prefix is not an option what about only exporting the traits in the prelude with the `*Prelude` suffix as suggested [here](https://github.com/aturon/rfcs/blob/eb5e3636557bad0d07ec71eaa5dcd5ff1a0554e2/active/0000-conventions-galore.md)? and establish as a convention to structure imports like this:
```rust
use async_graphql::{self as gql, prelude::*};

#[gql::SimpleObject]
struct MyObject {
    id: gql::ID,
    any: gql::Any,
    // ...
}
```

sunli829:
How about this?

```rust
use async_graphql::prelude::*;

#[GqlSimpleObject]
struct MyObject {
    id: GqlID, // GQL_ID, GID
    any: GqlAny,
    // ...
}
```

nicolaiunrein:
Besides the fact that it looks like Java 😆 I really don't think having two names point to the same type is a good idea.

phated:
We shouldn't have 2 names, but we also cannot export very generic names from the prelude. How do you want to do both without these changes?

nicolaiunrein:
> We shouldn't have 2 names, but we also cannot export very generic names from the prelude. How do you want to do both without these changes?

The only solution then is not to include the types in the prelude. And only have traits there which are already generically implemented but required to be in the namespace.

I think renaming certain types like `Enum` to `GqlEnum` or even `GraphqlEnum` or `GraphQLEnum` may be justified because it really is a "GraphQL Enum" as defined in the spec after all. I am more concerned with types like Result, Data, Error etc.

phated:
I encountered the problem with Result last night while trying to do some upgrades. This is why I support the prefixed names.

IcanDivideBy0:
@sunli829 
> I'm so sorry, this is my mistake, I did not think about it yesterday, and here I solemnly apologize to @IcanDivideBy0. 
It's all fine really, I should have thought of it too :)

I like the `use async_graphql as gql` version, I think if we put it everywhere in tests and documentation, most people will use it this way, hence the need for a prelude will disolve.
```rust
// I don't have strong opinion on this:
use async_graphql::prelude::*;
use async_graphql as gql;

// then...
#[GqlObject]
#[gql::Object] // I find this one more readable
```

That being said I think we should keep some of the prefixed version: `GqlResult` `GqlError` `GqlValue`

As you all know, those three are really common names in rust world, and while refactoring the code I found many place that required namespace juggling due name clashes.

IcanDivideBy0:
`use async_graphql as gql;` can even be completely removed if users declares the dependency as follows:
```
gql = { package = "async_graphql", version = "..." }
```

sunli829:
> That being said I think we should keep some of the prefixed version: GqlResult GqlError GqlValue. As you all know, those three are really common names in rust world, and while refactoring the code I found many place that required namespace juggling due name clashes around these three particular types.

I'm really happy to see you are not frustrated by this. 🍺
I support the naming of `GqlResult` and `GqlError`, but I especially don't like `GqlFieldResult`. I cried when I saw this in the morning.

nicolaiunrein:
@IcanDivideBy0 I agree mostly on what you are saying. Two points though:

1. Leaving the the prefix on eg. `GqlError` would result in using `gql::GqlError`. I don‘t see the point here. 

2. The prelude might still be neccessary for genericly implemented traits. Although I don‘t know how many there are.

sunli829:
But I remember I read in the rust book that `IoError` is not recommended, but `io::Error` should be used instead.

IcanDivideBy0:
@nicolaiunrein 
> GqlError would result in using gql::GqlError
You're right, that's ugly too :/

sunli829:
Like `serde::Result`, `actix_web::Result`, they are not bothered by the problem of naming Result.

nicolaiunrein:
> But I remember I read in the rust book that `IoError` is not recommended, but `io::Error` should be used instead.

That‘s what I‘m saying.

IcanDivideBy0:
Damn, naming things is HARD 😂

sunli829:
I also think prelude is necessary. In fact, the most important thing is to import the necessary traits.

sunli829:
In fact, unless writing test code, I never import *

sunli829:
The most fundamental reason is that the name gql is occupied by others. 😂

IcanDivideBy0:
Up until now `Guard` and `DataSource` are the two only traits I've had to import in my code base, and those two are not really general purpose

sunli829:
I just thought ErrorExtensions, ResultExt will be more `commonly` used.

IcanDivideBy0:
Well I just tried to rename to `gql` in Cargo.toml and this breaks all macros:
```
   |
18 | #[gql::Object]
   | ^^^^^^^^^^^^^^ use of undeclared type or module `async_graphql`
   |
```
but it works fine without renaming in cargo.toml and using `use async_graphql as gql;`

nicolaiunrein:
> I just thought `ErrorExtensions`, `ResultExt` will be more `commonly` used.

If that's the only two traits then I think it's not worth a prelude. I don't know how many people are actually using those. In Apollo there is a section in the book but I think there is trend in GraphQL towards returning result-enums much like in Rust to get typed errors. In fact that's what I started doing after implementing `ErrorExtensions` :laughing:

sunli829:
Let's put this question aside for now. 😁

PR #80, opened 2020-05-12
Allow for crate renaming in Cargo.toml

IcanDivideBy0:
Macros will now correctly expands when users rename the crate in cargo.toml

### DIFF
--- a/async-graphql-derive/Cargo.toml
+++ b/async-graphql-derive/Cargo.toml
@@ -21,3 +21,4 @@ proc-macro2 = "1.0.6"
 syn = { version = "1.0.13", features = ["full"] }
 quote = "1.0.3"
 Inflector = "0.11.4"
+proc-macro-crate = "0.1.4"
--- a/async-graphql-derive/src/utils.rs
+++ b/async-graphql-derive/src/utils.rs
@@ -1,5 +1,6 @@
 use async_graphql_parser::Value;
 use proc_macro2::{Span, TokenStream};
+use proc_macro_crate::crate_name;
 use quote::quote;
 use syn::{Attribute, Error, Expr, Ident, Lit, Meta, MetaList, NestedMeta, Result};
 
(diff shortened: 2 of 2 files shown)


IcanDivideBy0:
Macros will now correctly expands when users rename the crate in cargo.toml

declaring dependency like so now works
```
gql = { package = "async-graphql", version = "1.11.3" }
```

There's still an issue, but I don't know what's happening there.

This works:
```rust
use gql::FieldResult;

#[gql::Object]
impl User {
    async fn email(&self) -> Fieldresult<&str> {
        Ok("foo")
    }
}
```

This does not:
```rust
#[gql::Object]
impl User {
    async fn email(&self) -> gql::Fieldresult<&str> {
        Ok("foo")
    }
}
```
```
error: Invalid type
  --> src/api/models/user.rs:32:30
   |
32 |     async fn email(&self) -> gql::FieldResult<&str> {
   |                              ^^^^^^^^^^^^^^^^^^^^^^
```

sunli829:
Great. 😁

sunli829:
> async fn email(&self) -> gql::FieldResult<&str>

Let me solve this problem, but first tell me how to rename it in Cargo.toml. 😁

IcanDivideBy0:
Just declare your dependency like so:
```
gql = { package = "async-graphql", version = "1.11.3" }
```
You can use the same name in patch section to use your local version:
```
[patch.crates-io]
gql = { package = "async-graphql", path = "../../async-graphql" }
```

sunli829:
Thanks, I learned new knowledge again. 🍺

sunli829:
But I can start to solve this problem tomorrow. Because COVID-19 I haven't gone out for a long time to drink with my friends, and today is my first time in 4 months.

IcanDivideBy0:
Have fun! There's no pressure on this :D

PR #92, opened 2020-05-16
Fix compilation errors

c410-f3r:
`cargo build --no--default-features` doesn't compile due to missing guards.

1. Fix conditional features
2. Add CI step to avoid future breakages


sunli829:
Thank you. 😄

Issue #66, opened 2020-05-10
Add test code for introspection.


sunli829:

krevativ:
Can you elaborate more on which parts should be tested?

sunli829:
Sorry, I did not describe clearly.
GraphQL allows querying the definition of Schema, called introspection. Here is a description of the introspection. http://spec.graphql.org/June2018/#sec-Introspection
Now need to write some test code, define Schema with `async-graphql`, and then query it to see whether the correct output. 😁

sunli829:
I'll write a boilerplate code today, it shows how to do this test.

phated:
More detail also exists at https://github.com/graphql/graphql-spec#introspection, include a link to the introspection tests that the spec team use.

sunli829:
@krevativ  I added a boilerplate test code. 😁

krevativ:
Seems to be straight forward. Now I just need to understand all possible states :sweat_smile: But I'm new to GraphQL and it's a good way to learn more about introspection. Ok I do it and try to stay  close to the spec.

@sunli829 thanks for the boilerplate!

krevativ:
Ok I've looked more deeply into the spec. So far, these are the tests I want to write:

```
pub async fn test_introspection_simple_object() {}
pub async fn test_introspection_object() {}
pub async fn test_introspection_input_object() {}
pub async fn test_introspection_documentation() {}
pub async fn test_introspection_depraction() {}
pub async fn test_introspection_typename() {}
pub async fn test_introspection_schema() {}
pub async fn test_introspection_enum() {}
pub async fn test_introspection_interfaces() {}
pub async fn test_introspection_union() {}
pub async fn test_introspection_mutation() {}
```

What do you think?

I've also some questions:

1. What about the fields `interfaces`, `possibleTypes`, `enumValues`? Do they suppose to work? Because they don't or I do something wrong. (I think also the `inputFields`, but I haven't  tested it yet). 

2. What about the `__schema` field? It is a part of introspection spec. Should it be tested too?

3. I think subscriptions should also be tested. But I don't know how? :)

sunli829:
> 1. What about the fields interfaces, possibleTypes, enumValues? Do they suppose to work? Because they don't or I do something wrong. (I think also the inputFields, but I haven't tested it yet).

Yes, these need to be tested. For example, the specification defines that `interfaces` must return null, and` async-graphql` returns `[]`, `" "`, which is not correct.

[...]

PR #93, opened 2020-05-17
Introspection tests

krevativ:
#66 This pull request adds introspection  tests.

Here is my first shot on introspection tests. What do you think?

There are few things I want to mention here:

1. Now, introspection takes only the first line of the doc comment. I think it should be possible to use multiline doc comments too?

2. I couldn't understand how to add a description to a custom scalar. `#[Scalar(desc = "...")]` seems not to be implemented? In general, I think, it is often a bit confusing where to put the description. Maybe there should be more optional ways to do it and some kind of priority?

3. I took github api as a reference. And on github using similar test like `test_introspection_mutation`, for a query like `fields { type { kind name } }`, you would get `{ "kind": "OBJECT", "name": "SimpleObject" }` as return value. But you get `"kind": "NON_NULL", "name": null`. I'm not sure if this is an error?

4. Regarding the schema I don't know what to do. The result order is not the same on each run. You can try it out by commenting out the `test_introspection_schema`. `directives` and `types` have always a different order. So I can't just match on json format, I need to match individual fields. That's a lot of work, so I wanted to ask first if this supposed to be? Another question is, how detailed should this test be? Because I could imagine it to be pretty big? :smile: Or should I just match over a full query? 

TO-DO:
- [ ] Add schema tests


sunli829:
It's so exciting, I think it's very good test code.

> 1. Now, introspection takes only the first line of the doc comment. I think it should be possible to use multiline doc comments too?

Yes, I think this is a bug. 😁

> 2. I couldn't understand how to add a description to a custom scalar. #[Scalar(desc = "...")] seems not to be implemented? In general, I think, it is often a bit confusing where to put the description. Maybe there should be more optional ways to do it and some kind of priority?

I forgot to add the desc attribute to the scalar.

> 3. I took github api as a reference. And on github using similar test like test_introspection_mutation, for a query like fields { type { kind name } }, you would get { "kind": "OBJECT", "name": "SimpleObject" } as return value. But you get "kind": "NON_NULL", "name": null. I'm not sure if this is an error?

I will confirm the github specification again.

> 4. Regarding the schema I don't know what to do. The result order is not the same on each run. You can try it out by commenting out the test_introspection_schema. directives and types have always a different order. So I can't just match on json format, I need to match individual fields. That's a lot of work, so I wanted to ask first if this supposed to be? Another question is, how detailed should this test be? Because I could imagine it to be pretty big? 😄 Or should I just match over a full query?

The current order is indeed not stable, I think I sort the output by name, what do you think?

--------------

Thank you again for your contribution to async-graphql. 😁

krevativ:
@sunli829 you welcome :)

> Scalar description definition is different from other types, it is actually specified by ScalarType::description.

You are totally right! I've missed that. But this is what I mean by more options. Wouldn't it be nice to have also the proc_macro option? It would be more consistent with other types.

> I sorted __Schema.types and __Schema.directives by name.

Nice! Ok I will add schema tests a bit later then.

sunli829:
> You are totally right! I've missed that. But this is what I mean by more options. Wouldn't it be nice to have also the proc_macro option? It would be more consistent with other types.

I think your suggestion is very reasonable, so I decided to put the name and description of the scalar on the properties of the procedure macro to make it more consistent with other types.

PR #100, opened 2020-05-20
Stream connection

### LINKED ISSUE #98: Implement DataSource for Stream<T>
I'm planing to add a generic implementation of `DataSource` for `Stream`s in the same fashion it's done for slice. I'm using mongodb on m project, which returns a `Stream` of documents.

I think it's kind of wasteful to collect all stream results into a vec to then apply `query_operation` filter on this vec. I don't know about other database systems, but I think most of them are using stream also?

Does anyone have thoughts on this?

IcanDivideBy0:
fixes #98

### DIFF
--- a/src/types/connection/stream.rs
+++ b/src/types/connection/stream.rs
@@ -11,15 +11,13 @@ where
     stream: Pin<Box<dyn Stream<Item = (Cursor, E, T)> + Send + Sync + 'a>>,
 }
 
-impl<'a, T, E> StreamDataSource<'a, T, E>
+impl<'a, S, T, E> From<S> for StreamDataSource<'a, T, E>
 where
+    S: Stream<Item = (Cursor, E, T)> + Send + Sync + 'a,
     T: Send + Sync + 'a,
     E: ObjectType + Send + Sync + 'a,
 {
-    pub fn new<S>(stream: S) -> Self
-    where
-        S: Stream<Item = (Cursor, E, T)> + Send + Sync + 'a,
-    {
+    fn from(stream: S) -> Self {
         StreamDataSource {
             stream: Box::pin(stream),
         }
--- a/src/types/connection/stream.rs
+++ b/src/types/connection/stream.rs
@@ -1,8 +1,62 @@
 use crate::types::connection::{EmptyEdgeFields, QueryOperation};
 use crate::{Connection, Context, Cursor, DataSource, FieldResult, ObjectType};
 use futures::{Stream, StreamExt};
+use std::collections::VecDeque;
 use std::pin::Pin;
 
+/// Utility struct to convert a Stream to a Connection
+///
+/// # Examples
+///
+/// ```rust
+/// use async_graphql::*;
+/// use byteorder::{ReadBytesExt, BE};
+/// use futures::StreamExt;
+///
+/// struct QueryRoot;
+///
+/// #[Object]
+/// impl QueryRoot {
+///     async fn stream_connection(&self, ctx: &Context<'_>,
+///         after: Option<Cursor>,
+///         before: Option<Cursor>,
+///         first: Option<i32>,
+///         last: Option<i32>
+///     ) -> FieldResult<Connection<&str>> {
+///         let edges_stream = futures::stream::iter(vec!["a", "b", "c", "d", "e", "f"])
+///             .map(|node| {
+///                 let cursor: Cursor = node.to_owned().into();
+///                 (cursor, EmptyEdgeFields, node)
+///             });
+///
+///         let mut source: StreamDataSource<_> = edges_stream.into();
+///         source.query(ctx, after, before, first, last).await
+///     }
+/// }
+///
+/// #[async_std::main]
+/// async fn main() {
+///     let schema = Schema::new(QueryRoot, EmptyMutation, EmptySubscription);
+///
+///     assert_eq!(schema.execute("{ streamConnection(first: 2) { edges { node } } }").await.unwrap().data, serde_json::json!({
+///         "streamConnection": {
+///             "edges": [
+///                 { "node": "a" },
+///                 { "node": "b" }
+///             ]
+///         },
+///     }));
+///
+///     assert_eq!(schema.execute("{ streamConnection(last: 2) { edges { node } } }").await.unwrap().data, serde_json::json!({
+///         "streamConnection": {
+///             "edges": [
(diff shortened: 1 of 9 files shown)


IcanDivideBy0:
@phated @sunli829 tell me if its ok on the main idea, I'll add documentation tomorrow I think

src/types/connection/stream.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+use std::collections::VecDeque;
+struct State<E: Send + Sync> {
+    has_seen_before: bool,
+    has_prev_page: bool,
+    has_next_page: bool,
+    edges: VecDeque<(Cursor, EmptyEdgeFields, E)>,
+}
```
IcanDivideBy0, line 33: Not really happy with this, but I can't get a better idea :/

tests/connection.rs
```diff
@@ -0,0 +1,282 @@
+pub mod stream {
+    use async_graphql::*;
+    use serde_json::{json, Value::Null};
```
sunli829, line 3:
```suggestion
    use serde_json::json;
```

tests/connection.rs
```diff
@@ -0,0 +1,282 @@
...
+                    "edges": edges,
+                },
+            })
+        );
+    }
+
+    #[async_std::test]
+    pub async fn test_connection_stream_none() {
+        do_test(
+            json!({ "after": Null, "before": Null, "first": Null, "last": Null }),
```
sunli829, line 88:
```suggestion
            json!({ "after": null, "before": null, "first": null, "last": null }),
```

tests/connection.rs
```diff
@@ -0,0 +1,282 @@
...
+            let stream = futures::stream::iter(vec![
+                "a".to_owned(),
+                "b".to_owned(),
+                "c".to_owned(),
+                "d".to_owned(),
+                "e".to_owned(),
+                "f".to_owned(),
+            ]);
+
+            let mut source = StreamDataSource::new(stream, |s: &String| {
```
phated, line 26: Is this iterating the entire stream to gather all cursors? If so, it seems inefficient that we'd iterate it twice

IcanDivideBy0: No, the stream is just mapped with this closure:
https://github.com/async-graphql/async-graphql/pull/100/files#diff-c6747b5500cb9b8d48c79e5d07e7daa9R18

src/types/connection/stream.rs
```diff
@@ -0,0 +1,144 @@
...
+{
+    pub fn new<S>(stream: S) -> Self
+    where
+        S: Stream<Item = (Cursor, E, T)> + Send + Sync + 'a,
+    {
+        StreamDataSource {
+            stream: Box::pin(stream),
+        }
+    }
+}
```
IcanDivideBy0, line 27: I've removed the mapping function for cursor, `StreamDataSource` now takes a `Stream<Item = (Cursor, E, T)>`. This gives more flexibility to users and is easy enouth for anyone to handle it in their codebase. Also gives the ability to control the EdgeFieldsObj. (see test)

IcanDivideBy0: I think we should adopt the same strategy for the slice implementation.

IcanDivideBy0:
Just some more thoughts: maybe we should remove the slice implementation in favor of anything that implement iterator ?
Then web can convert iterator into stream easily and use the same code for both ?

src/types/connection/stream.rs
sunli829: I agree. 😁

IcanDivideBy0: removing the closure, we can now have `impl From<Stream> for StreamDataSource`, which is nice :)

sunli829:
I fixed the problem that the master branch could not be compiled.

sunli829:
@IcanDivideBy0  Can it be merged? 😁

IcanDivideBy0:
@sunli829 yep, just wait for tests!

The current DataSource impl for slice is kind of whacky :
- cursor should not be used a base64 slice indices because push_front in the slice shifts all cursors to the next node, hence nodes may have different cursor based on their position
- edges fields are locked to `EmptyEdgeFields`
I think we should remove it but it would introduce breaking change...

IcanDivideBy0:
@sunli829 please don't forget this is breaking change PR when releasing since `query_operation` now takes `&mut self`

IcanDivideBy0:
wait, I think we can remove the `Sync` requirement for `DataSource`, which will simplify some code...

IcanDivideBy0:
That's neat !
We can now query directly over a `BoxStream` :
```rust
async fn stream_connection(
    &self,
    ctx: &Context<'_>,
    after: Option<Cursor>,
    before: Option<Cursor>,
    first: Option<i32>,
    last: Option<i32>,
) -> FieldResult<Connection<String>> {
    futures::stream::iter(vec!["a", "b", "c", "d", "e", "f"])
    .map(|node| {
        let cursor: Cursor = node.clone().into();
        (cursor, EmptyEdgeFields, node)
    })
    .boxed()
    .query(ctx, after, before, first, last)
    .await
}
```

\o/

sunli829:
The slice datasource is just for testing, and I don't believe anyone would actually use it. 😀

sunli829:
> That's neat !
> 
> We can now query directly over a `BoxStream` :
> 
> ```rust
> 
> async fn stream_connection(
> 
>     &self,
> 
>     ctx: &Context<'_>,
> 
>     after: Option<Cursor>,
> 
>     before: Option<Cursor>,
> 
>     first: Option<i32>,
> 
>     last: Option<i32>,
> 
> ) -> FieldResult<Connection<String>> {
> 
>     let mut stream = futures::stream::iter(vec!["a", "b", "c", "d", "e", "f"])
> 
>         .map(|node| {
> 
>             let cursor: Cursor = node.clone().into();
> 
>             (cursor, EmptyEdgeFields, node)
> 
>         })
> 
>         .boxed();
> 
>     
> 
>     stream.query(ctx, after, before, first, last).await
> 
> }
> 
> ```
> 
> 
> 
> \o/


Amazingly easy to use!😂

sunli829:
When you think there is no problem, you can merge them yourself. 
Thank you very much!😁

IcanDivideBy0:
One more improvement and we're good to go:
I think a loop would be more efficient since I can break out of it anytime even if the stream is not terminated.

PR #105, opened 2020-05-24
New data source

sunli829:
I rewrote the code of DatsSource.
The DataSource implemented for Stream has been temporarily removed, and this modification will be added after everyone agrees.


src/types/connection/connection_type.rs
```diff
@@ -1,97 +1,136 @@
...
-    pub nodes: Vec<(Cursor, E, T)>,
+    pub(crate) nodes: Vec<Record<String, E, T>>,
 }
 
-impl<T, E: ObjectType + Send> Connection<T, E> {
-    /// Create a connection object.
-    pub fn new(
-        total_count: Option<usize>,
+impl<T: OutputValueType + Send, E: ObjectType + Send> Connection<T, E> {
+    pub fn empty() -> Self {
```
IcanDivideBy0, line 57: Givent the empty values, I guess `Connection` and `PageInfo` could derive `Default`?

src/types/connection/cursor.rs
```diff
@@ -1,60 +1,27 @@
...
+use std::fmt::Display;
 
-/// Cursor scalar
+/// Cursor type
 ///
 /// A custom scalar that serializes as a string.
 /// https://relay.dev/graphql/connections.htm#sec-Cursor
-#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Debug)]
-pub struct Cursor(String);
+pub trait CursorType: Sized {
```
IcanDivideBy0, line 8: I think web should impl `CursorType` for `String` and `ID` as well

src/base.rs
```diff
@@ -38,14 +38,19 @@ pub trait Type {
...
     fn from_global_id(id: ID) -> Option<ID> {
-        let v: Vec<&str> = id.splitn(2, ':').collect();
-        if v.len() != 2 {
-            return None;
-        }
-        if v[0] != Self::type_name() {
-            return None;
-        }
-        Some(v[1].to_string().into())
+        base64::decode(id.as_str())
```
phated, line 41: This is a weird pattern to force people into. If you structure data properly, this isn't needed because you can do a trip to the database to find the type. You shouldn't be relying on types sent from the client.

src/types/connection/connection_type.rs
```diff
@@ -1,97 +1,136 @@
...
 
+/// The record type output by the data source
+pub struct Record<C, E, T> {
+    cursor: C,
+    edge: E,
+    element: T,
+}
+
+impl<C, E, T> Record<C, E, T> {
+    pub fn new(cursor: C, edge: E, element: T) -> Self {
```
phated, line 23: This is great! 👍 I was going to implement this for my own stuff.

src/types/connection/connection_type.rs
```diff
@@ -1,97 +1,136 @@
...
+        Self {
+            cursor,
+            edge,
+            element,
+        }
+    }
+}
+
+impl<C, T> Record<C, EmptyEdgeFields, T> {
+    pub fn new_without_edge_fields(cursor: C, element: T) -> Self {
```
phated, line 33: I think we should make `new` be without EdgeFields and the make this `new_with_edge_fields`. The right defaults are important.

src/types/connection/connection_type.rs
sunli829: I tried it and failed to compile it. 🙁

src/types/connection/connection_type.rs
```diff
@@ -1,97 +1,136 @@
...
 };
 use async_graphql_parser::query::Field;
+use futures::{Stream, TryStreamExt};
 use indexmap::map::IndexMap;
 use inflector::Inflector;
 use itertools::Itertools;
 use std::borrow::Cow;
 
+/// The record type output by the data source
+pub struct Record<C, E, T> {
```
phated, line 16: If we want to map to the relay spec, I believe this should be named `Edge`

src/types/connection/connection_type.rs
```diff
@@ -1,97 +1,136 @@
...
+    cursor: C,
+    edge: E,
```
phated, line 18: This would then be named `additional_fields` (or something similar)

src/types/connection/connection_type.rs
```diff
@@ -1,97 +1,136 @@
...
+    element: T,
```
phated, line 19: And this would be named `node`

src/types/connection/connection_type.rs
phated: If we change the name of my suggestion above, we could call this `new_with_additional_fields`.

src/types/connection/connection_type.rs
```diff
@@ -1,97 +1,136 @@
...
+                has_next_page: false,
+                start_cursor: None,
+                end_cursor: None,
+            },
+            nodes: Vec::new(),
+        }
+    }
+
+    pub fn new<C>(
+        nodes: Vec<Record<C, E, T>>,
```
phated, line 71: Following the recommended naming, this could be named `edges` which is a `Vec<Edge<C, E, T>>`

src/types/connection/connection_type.rs
```diff
@@ -1,97 +1,136 @@
...
-    ) -> Self {
+        total_count: Option<usize>,
+    ) -> Connection<T, E>
+    where
+        C: CursorType,
+    {
+        let nodes = nodes
+            .into_iter()
+            .map(|record| Record {
+                cursor: record.cursor.encode_cursor(),
```
phated, line 82: I'm wondering if this needs to be similar to `try_into` since it should be allowed to fail.

src/types/connection/connection_type.rs
```diff
@@ -1,97 +1,136 @@
...
+            })
+            .collect_vec();
         Connection {
             total_count,
             page_info: PageInfo {
                 has_previous_page,
                 has_next_page,
-                start_cursor: nodes.first().map(|(cursor, _, _)| cursor.clone()),
-                end_cursor: nodes.last().map(|(cursor, _, _)| cursor.clone()),
+                start_cursor: nodes.first().map(|record| record.cursor.clone()),
```
phated, line 92: Is there a reason `Cursor` can't derive `Copy`?

src/types/connection/connection_type.rs
```diff
@@ -180,16 +219,36 @@ impl<T: OutputValueType + Send + Sync, E: ObjectType + Sync + Send> ObjectType
...
+                        extra_type: &record.edge,
+                        node: &record.element,
+                    })
+                })
+                .collect_vec();
+            return OutputValueType::resolve(&edges, &ctx_obj, ctx.item).await;
         } else if ctx.name.node == "totalCount" {
             let ctx_obj = ctx.with_selection_set(&ctx.selection_set);
-            return OutputValueType::resolve(&self.total_count().await, &ctx_obj, ctx.item).await;
+            return OutputValueType::resolve(
```
phated, line 239: Since `total_count` isn't required by the spec, is this allowed to be excluded by an implementation?

src/types/connection/cursor.rs
```diff
@@ -1,60 +1,27 @@
...
-    fn deref(&self) -> &Self::Target {
-        &self.0
-    }
-}
-
-impl DerefMut for Cursor {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.0
-    }
+    fn encode_cursor(self) -> String;
```
phated, line 13: I think encoding should be fallible too.

src/types/connection/cursor.rs
phated: I agree, but we shouldn't force base64 encoding/decoding on them. There's not a forced rule to implement base64.

I use UUIDs that are "opaque enough" for my system.

src/types/connection/cursor.rs
```diff
@@ -1,60 +1,27 @@
...
-        }
+    fn decode_cursor(s: &str) -> Result<Self, Self::DecodeError> {
+        let data = base64::decode(s)?;
+        Ok(data.as_slice().read_u32::<BE>()? as usize)
     }
 
-    fn to_json(&self) -> Result<serde_json::Value> {
-        Ok(self.0.to_string().into())
+    fn encode_cursor(self) -> String {
+        base64::encode((self as u32).to_be_bytes())
```
phated, line 25: This should be fallible.

src/types/connection/page_info.rs
```diff
@@ -1,20 +1,17 @@
...
     pub has_previous_page: bool,
 
-    #[field(desc = "When paginating forwards, are there more items?")]
+    /// When paginating forwards, are there more items?
     pub has_next_page: bool,
 
-    #[field(desc = "When paginating backwards, the cursor to continue.")]
-    pub start_cursor: Option<Cursor>,
+    /// When paginating backwards, the cursor to continue.
+    pub start_cursor: Option<String>,
```
phated, line 13: `Option<impl CursorType>`?

src/types/connection/page_info.rs
```diff
@@ -1,20 +1,17 @@
...
 
-    #[field(desc = "When paginating forwards, the cursor to continue.")]
-    pub end_cursor: Option<Cursor>,
+    /// When paginating forwards, the cursor to continue.
+    pub end_cursor: Option<String>,
```
phated, line 16: `Option<impl CursorType>`?

src/base.rs
sunli829: I removed `Type::global_id` and `Type::from_global_id`. 🙂

src/types/connection/page_info.rs
sunli829: This requires `PageInfo<C>`, but it's an internal type, so `String` might be more convenient.

src/types/connection/connection_type.rs
IcanDivideBy0: Strange, looks like the compiler want C, T and E to implement default as well, even if they're only used in a Vec and Option, which impl Default whatever the type :/

src/types/connection/connection.rs
```diff
@@ -1,100 +1,115 @@
...
+                start_cursor: None,
+                end_cursor: None,
+            },
+            edges: Vec::new(),
+        }
+    }
+
     pub fn new(
-        total_count: Option<usize>,
+        nodes: Vec<Edge<C, T, E>>,
```
IcanDivideBy0, line 49: this is a vec of `edges`, not `nodes`

src/types/connection/edge.rs
```diff
@@ -1,39 +1,45 @@
...
 use std::borrow::Cow;
 
-pub struct Edge<'a, T, E> {
-    pub cursor: &'a str,
-    pub node: &'a T,
-    pub extra_type: &'a E,
+/// The edge type output by the data source
+pub struct Edge<C, T, E> {
+    pub cursor: C,
+    pub element: T,
```
IcanDivideBy0, line 14: An element inside an edge is named a `node`

sunli829:
> This is really looking good. Other than the naming comments from @IcanDivideBy0, I'm giving 👍
> 
> Should we make `EdgeFieldsType` an `Option` to remove `EmptyEdgeFields` from the API? Now that we have an API for `new_with_additional_fields`, users wouldn't need to operate on the Some/None.

This may not be removable, maybe I have not found a way. 😀

src/types/connection/connection_type.rs
sunli829: If we remove the `totalCount`, when the user wants this field, we have not provided a way to add any field to the Connection. 😂

phated: @sunli829 if we don't make this optional, we cannot support streams that end between a `cursor` and `limit` and you force the entire stream to be consumed. In large data sets, that fails.

phated:
@sunli829 do you want me to try making `EdgeFieldsType` an `Option`?

sunli829:
> @sunli829 do you want me to try making `EdgeFieldsType` an `Option`?

Yes, thank you, and I want to remove the `totalCount` on the `Connection` and add additional fields to the Connection. 😁

sunli829:
> > @sunli829 do you want me to try making `EdgeFieldsType` an `Option`?
> 
> Yes, thank you, and I want to remove the `totalCount` on the `Connection` and add additional fields to the Connection. 😁

I am going to solve this problem now. There are potential security issues here, and I found a way to avoid using unsafe code.

https://github.com/async-graphql/async-graphql/blob/2bb2c25777ebc91cf976dd37e8eefec7d720ed45/src/subscription/connection.rs#L132

phated:
👍 I will let you work on it! 🍻

sunli829:
Is it impossible to make `EdgeFieldsType` into `Option`? @phated  😁

phated:
I didn't work on it because I thought you were trying. Would you like me to work on it?

sunli829:
> I didn't work on it because I thought you were trying. Would you like me to work on it? 

Yes. 😅

sunli829:
I don't know how to do this. Maybe you can help me with it. 😁

sunli829:
We may need to remove the generic parameter `E` on `Connection` to do this.

phated:
@sunli829 I don't see an easy way to support this without a lot of refactor.

sunli829:
> @sunli829 I don't see an easy way to support this without a lot of refactor.

Only trait objects can be used, which will reduce performance and feel unnecessary.

phated:
I honestly don't understand the benefit of these extra fields, but I just want this feature finished so I can use it.

sunli829:
> @sunli829 I don't see an easy way to support this without a lot of refactor.



> I honestly don't understand the benefit of these extra fields, but I just want this feature finished so I can use it.

Agree.

sunli829:
Let's leave it at that, I'm going to add the Connection's additional field.

src/types/connection/edge.rs
```diff
@@ -10,27 +10,29 @@ use std::borrow::Cow;
...
-    pub element: T,
-    pub additional_fields: E,
+    pub(crate) cursor: C,
+    pub(crate) node: T,
+    pub(crate) additional_fields: E,
 }
 
 impl<C, T, E> Edge<C, T, E> {
+    /// Create a new edge, it can have some additional fields.
     pub fn new_with_additional_fields(cursor: C, element: T, additional_fields: E) -> Self {
```
IcanDivideBy0, line 20: maybe also rename the parameter `node` ? I think it makes things more clear to anyone who have already worked with relay conection

phated:
@sunli829 I added this to my project locally and it is very cool! I'm excited to see @IcanDivideBy0 re-introduce his Stream support

sunli829:
> @sunli829 I added this to my project locally and it is very cool! I'm excited to see @IcanDivideBy0 re-introduce his Stream support

The current implementation does not conflict with StreamDataSource and can be easily added again. @IcanDivideBy0

IcanDivideBy0:
@phated @sunli829 I'll take a look a this today!

mwilliammyers:
I don’t know how I didn’t catch this before but according to the [Rust guidelines for APIs](https://rust-lang.github.io/api-guidelines/naming.html?highlight=with_#casing-conforms-to-rfc-430-c-case), it should be `with_additional_fields` not `new_with_additional_fields`. I know it’s small but, @sunli829, you said you appreciate clean APIs 🙃

Is it worth changing?

phated:
@mwilliammyers seems good. Want to open an issue? I think we can keep both around for awhile since it doesn't hurt anything.

sunli829:
> I don’t know how I didn’t catch this before but according to the [Rust guidelines for APIs](https://rust-lang.github.io/api-guidelines/naming.html?highlight=with_#casing-conforms-to-rfc-430-c-case), it should be `with_additional_fields` not `new_with_additional_fields`. I know it’s small but, @sunli829, you said you appreciate clean APIs 🙃
> 
> Is it worth changing those?



> I don’t know how I didn’t catch this before but according to the [Rust guidelines for APIs](https://rust-lang.github.io/api-guidelines/naming.html?highlight=with_#casing-conforms-to-rfc-430-c-case), it should be `with_additional_fields` not `new_with_additional_fields`. I know it’s small but, @sunli829, you said you appreciate clean APIs 🙃
> 
> Is it worth changing those?

This is a good suggestion, let me try to make some changes.

sunli829:
At the beginning, `with_additional_fields` was not provided because it usually seems to be an option, but in fact it is a required option.

```rust
Connection :: new(...).with_additional_fields (...);
```

If `with_additional_fields` is missing, it will fail to compile,

sunli829:
Do you just want to change the name of this function?

phated:
@sunli829 yes it is just to change the name because of Rust guideline recommendation.

IcanDivideBy0:
Ok, now that I've spend some hours trying to get thing working for Stream, I'll share here where I've come...

The new CursorType is not really easy to work with, mostly because of the Sync requirement. I've managed to almost completely remove the Sync everywhere from Schema, Resolver, ObjectType impls etc but got stuck at some point with Data. I still think this can be achieved but it's getting too far beyond my comprehension of the crate as a whole :/
(If someone can point me why we cannot remove Sync from almost the whole API, that'd be enlightening)

I think the most easy way would be to remove the CursorType trait which don't really add any value to async_graphql, forcing it as String everywhere it's used and let users encode / decode their Cursor in the `execute_query` function since it already return a `Result`. Also It feels a bit strange to return a Connection with the cursor type set to ID and getting a String in the schema for example.

That being said, one change that will be required anyways to implement DataSource for Stream is to change `query` and `execute_query` signatues to `&mut self`

sunli829:
Let me think for a moment. 😀

sunli829:
Can you tell me where you use stream datasource？

IcanDivideBy0:
I was trying to bring back
```rust
impl<C, T> DataSource for Pin<Box<Stream<(C, T)>>>
where
    C: CursorType,
    T: OutputValueType
```

sunli829:
Do you give the results of the MongoDB query to Stream DataSource?

sunli829:
> Ok, now that I've spend some hours trying to get thing working for Stream, I'll share here where I've come...
> 
> The new CursorType is not really easy to work with, mostly because of the Sync requirement. I've managed to almost completely remove the Sync everywhere from Schema, Resolver, ObjectType impls etc but got stuck at some point with Data. I still think this can be achieved but it's getting too far beyond my comprehension of the crate as a whole :/
> (If someone can point me why we cannot remove Sync from almost the whole API, that'd be enlightening)
> 
> I think the most easy way would be to remove the CursorType trait which don't really add any value to async_graphql, forcing it as String everywhere it's used and let users encode / decode their Cursor in the `execute_query` function since it already return a `Result`. Also It feels a bit strange to return a Connection with the cursor type set to ID and getting a String in the schema for example.
> 
> That being said, one change that will be required anyways to implement DataSource for Stream is to change `query` and `execute_query` signatues to `&mut self`

A few days ago I also wanted to remove Sync, but I failed. 😂

phated:
I also made my cursor a string. Should we change that?

sunli829:
> I also made my cursor a string. Should we change that?

I think this is not a problem. In most cases, the cursor should be a base64 encoded string.

phated:
@sunli829 we need to delete the cursor type. That's what is causing the Sync problems. Please make it only a String

sunli829:
Cursor is already implemented for the String type. What is the difference between removing Cursor and replacing it with a String？

phated:
I believe @IcanDivideBy0 already explained why it doesn't work but maybe we need to see his code?

IcanDivideBy0:
Ohhh my bad, I don't know why but I was absolutely convinced that a stream of Sync items **have** to be Sync... So I went straight up in this direction, trying to remov the Sync requirement on CursorType

So, in the end, this compile fine:
```rust
use crate::connection::{Connection, CursorType, DataSource, EmptyFields};
use crate::{Context, FieldResult, OutputValueType};
use async_graphql_derive::DataSource;
use futures::Stream;
use std::pin::Pin;

#[DataSource(internal)]
impl<'a, C, T> DataSource for Pin<Box<dyn Stream<Item = (C, T)> + Send + 'a>> // Not Sync !
where
    C: CursorType + Send + Sync + 'a,
    T: OutputValueType + Send + Sync + 'a,
{
    type CursorType = C;
    type NodeType = T;
    type ConnectionFieldsType = EmptyFields;
    type EdgeFieldsType = EmptyFields;

    async fn execute_query(
        &mut self,
        _ctx: &Context<'_>,
        after: Option<Self::CursorType>,
        before: Option<Self::CursorType>,
        first: Option<usize>,
        last: Option<usize>,
    ) -> FieldResult<
        Connection<
            Self::CursorType,
            Self::NodeType,
            Self::ConnectionFieldsType,
            Self::EdgeFieldsType,
        >,
    > {
        unimplemented!()
    }
}
```
Still a few gotchas tho
1. We need to get get a mutable ref to the stream to pull nodes out of it
2. Previously my first aim was to reduce memory usage by not having to pull a large stream into a vec to slice it right after completion. Since performances were pretty bad on previous implementation, I wonder if we really should provide this implementation in the crate.

phated:
I'm having a hard time conceptualizing how my database stream will be used with the Stream datasource since my return type is a GraphQL Union. The `&self` reference has been hard for me because I can't implement DataSource against a Rust `enum`, which we require for the Union macro—which required me to implement a dummy `struct` just to have a DataSource.

I was hoping to remove that `&self` reference completely and store data in the `Context`, but maybe that's the wrong way to think about this?

As for performance, I say we implement it with "bad" performance and then improve the performance once we can start using it in real scenarios. What do you think?

PR #107, opened 2020-05-25
Derive Serialize & Deserialize for Json

mwilliammyers:


phated:
Where are you serde'ing this? I believe the scalar implementation already serde's the inner type. Is your inner type missing derive serde?

mwilliammyers:
I must confess, I don't really know how to use the JSON scalar type 🙃 . But, I know when I added this patch to my local checkout and built against it, it got rid of compiler errors...

I am using it like this:

```rust
#[derive(Serialize, Deserialize, Clone, Debug)]
struct User {
    name: String,

    #[serde(default, skip_serializing_if = "Option::is_none")]
    metadata: Option<async_graphql::Json<serde_json::Value>>,
}
```

sunli829:
Let me take a look at the problem.

sunli829:
Thank you, this really works.

mwilliammyers:
Yeah, when I do:

```rust
#[derive(Serialize, Deserialize, Clone, Debug)]
struct User {
    metadata: serde_json::Value,
}

#[graphql::Object]
impl User {
    async fn metadata(&self) -> JsonValue {
        // it would be nice to implement it for `&JsonValue`
        self.metadata.clone()
    }
}
```

I get this:
```
the trait bound `serde_json::value::Value: async_graphql::base::OutputValueType` is not satisfied
the trait `async_graphql::base::OutputValueType` is not implemented for `serde_json::value::Value`
```

sunli829:
You should use `Json<serde_json::Value>` 😁, because `OutputValueType` is not implemented for `serde_json::Value`.

```rust
#[derive(Serialize, Deserialize, Clone, Debug)]
struct User {
    metadata: Json<serde_json::Value>,
}

```

Issue #115, opened 2020-05-27
Allow to differentiate between null and undefined


IcanDivideBy0:
Let's say I have a task object with an optional deadline
```rust
struct Task {
    text: String,
    deadline: Option<DateTime<Utc>>
}
```

Currently with my JS server I can update the task with some mutation:
```gql
mutation m1($id: ID!, $text: String!, $deadline: DateTime) {
  UpdateTask(id: $id, text: $text, deadline: $deadline ) {
    id
    text
    deadline
  }
}
```

The gql variable `$deadline` can takes 3 different types of values:
* a Date, which update the deadline
* `undefined`, which do not modify the deadline
* `null`, which clears the deadline

I can't find  a way to reproduce this with async_graphql, does anyone have an idea on how to handle this?

sunli829:
`Option<Option<DateTime>>`, is this ok? 😁

mwilliammyers:
Just to be clear:
- `None` -> `undefined`
- `Some(None)` -> `null`
- `Some(Some(x))` -> `x`

How would this look in graphql? Just an optional `DateTime`?

phated:
Does graphql actually allow undefined? I vaguely remember only seeing null in the spec.

mwilliammyers:
Don’t think so, http://spec.graphql.org/June2018/#sec-Value-Completion, but I know at least the Apollo Client/Server (in JS/node/TS) differentiate between null and undefined...

sunli829:
I was wrong. Graphql doesn't support this type.😂

mwilliammyers:
Waiiit a second, that won’t work because if a client passes `null` serde will treat that as `None` for the outer `Option`, which means `undefined` in our case, but we want it to mean `Some(None)` (null)

sunli829:
I think this can only add one parameter.

sunli829:
I'm thinking about adding something like this.

https://gqlgen.com/reference/changesets/

phated:
@sunli829 I'm not sure I understand why you would use "changesets"

I wonder if @IcanDivideBy0's mutation is just too generic. My gut feeling here would be that `setDeadline` should be a separate mutation than `updateTaskTitle`, but I'm unsure.

nicolaiunrein:
maybe one could construct something with `#[serde(deny_unknown_fields)]`

sunli829:
The first field `value1` is nested `Option`, so I can handle `undefined`, how does this look?

# Input object definition

```rust
[...]

PR #123, opened 2020-05-28
Add MaybeUndefined type

sunli829:
Allow to differentiate between null and undefined #115.


IcanDivideBy0:
Wow, I was about to try the double Option trick to see how it works and this comes out xD

sunli829:
Here is test code:

https://github.com/async-graphql/async-graphql/blob/8afe5427e109f9987ed9ed0ab40024777e461165/tests/maybe_undefined.rs#L4

src/types/maybe_undefined.rs
```diff
@@ -0,0 +1,89 @@
+use crate::{registry, InputValueResult, InputValueType, Type, Value};
+use std::borrow::Cow;
+
+/// Similar to `Option`, but it has three states, `undefined`, `null` and `x`.
+///
+/// Spec: https://spec.graphql.org/June2018/#sec-Null-Value
+#[allow(clippy::option_option)]
+pub struct MaybeUndefined<T>(Option<Option<T>>);
```
IcanDivideBy0, line 8: What about making it an enum ?
```rust
enum MaybeUndefined<T> {
    Undefined,
    Null,
    Value(T)
}
```
This would allow for easier pattern matching instead of `if is_null esle if is undefined else ...`

IcanDivideBy0:
If you're ok with going for the enum form and want to move on something else, I think I can complete the PR on my own now that you did all the hard work :D

sunli829:
> If you're ok with going for the enum form and want to move on something else, I think I can complete the PR on my own now that you did all the hard work :D

Good suggestion, I changed to enum.

sunli829:
> If you're ok with going for the enum form and want to move on something else, I think I can complete the PR on my own now that you did all the hard work :D

I just have nothing to do. 😂

tests/maybe_undefined.rs
```diff
@@ -0,0 +1,57 @@
...
+    struct MyInput {
+        value: MaybeUndefined<i32>,
+    }
+
+    struct Query;
+
+    #[Object]
+    impl Query {
+        async fn value1(&self, input: MaybeUndefined<i32>) -> i32 {
+            if input.is_null() {
```
IcanDivideBy0, line 15: Is this still works ?

sunli829: I submitted another one. 😁

IcanDivideBy0:
This looks good to me, perfectly fits my needs! :+1: 
@mwilliammyers @phated maybe you want to discuss a bit more about this feature before it gets merged?

Given the exemple in the issue, it could easily have been solved by introducing a new mutation, but i'm currently migrating my server (NodeJS graphql server, not apollo) which is already a pretty big task, I'm trying not to introduce API changes at the same time. I expect not to be alone in this case!

sunli829:
@IcanDivideBy0 I released version `1.14.2`.

Issue #112, opened 2020-05-26
Interface fields name camelCasing


Amareis:
```rust
#[Object]
impl PMChat {
    #[field(name = "created_at")]
    pub async fn created_at(&self) -> &String {
        &self.data.created_at
    }
}

#[Object]
impl Chat {
    #[field(name = "created_at")]
    pub async fn created_at(&self) -> &String {
        &self.data.created_at
    }
}

#[Interface(
    field(name = "created_at", type = "&String"),
)]
pub enum IChat {
    PMChat(PMChat),
    Chat(Chat),
}
```

And in resulting scheme interface field became camelCased:
```
interface IChat {
  createdAt: String!
}
```

There is no any compilation error nor way to disable name transformation.

Amareis:
I think interface fields should not be transformed, since it's already strings, so if I want to make it camelCased, I can just wrote it. Or maybe add different argument to `field` attribute for disabling it.

sunli829:
Let me think about how to solve it.

sunli829:
I'll add an alias field, because the interface query is forwarded via name and must be the same as the function name of the corresponding instance.

I'll do it tomorrow. I'm ready for bed now. 😊

Amareis:
May be also add disabling name mangling as a crate feature? I think it will be handy for projects with snake_case naming.

sunli829:
Async-graphql 1.13.3 already includes this change.

sunli829:
Here is the test code:

https://github.com/async-graphql/async-graphql/blob/afb2061392f5f54568a06ba464ac02593af08ee4/tests/interface.rs#L280

Amareis:
> Async-graphql 1.13.3 already includes this change.

Oh, you really awesome! Thanks you! I'll test it now.

Amareis:
Yep, it works. Maybe it should be mirrored in docs?

sunli829:
Can you help update the book? I will translate it into Chinese later. 😁

Amareis:
> Can you help update the book? I will translate it into Chinese later.

Ok, I can, but later. Currently I try to check performance of different server libraries for our production app. Can I contact you at email to discuss some questions about it?

sunli829:
> > Can you help update the book? I will translate it into Chinese later.
> 
> Ok, I can, but later. Currently I try to check performance of different server libraries for our production app. Can I contact you at email to discuss some questions about it?

Yes.

Amareis:
[...]

PR #124, opened 2020-05-28
Added docs for `Interface` field `method` argument

Amareis:
#112 completed


Amareis:
Version should be bumped after mege this, I think.

sunli829:
Thank you very much. 😁

PR #120, opened 2020-05-27
Add stability warning to README & book (ref #117)

phated:
@mwilliammyers How do you feel about this warning? Definitely looking for ways to improve it.

PR #128, opened 2020-05-31
Improve stability message based on #120 feedback

phated:
This brings in the suggestions from #120


sunli829:
Thank you very much. 😁

PR #150, opened 2020-06-05
Add possibility to use @defer/@stream on a fragment spread or an inline fragment

sync:
This is just to please the relay compiler, it doesn't stream the fragment... I am not sure how to do this at the moment.

PS: feel free to close this if it makes no sense

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -249,7 +249,20 @@ where
                 __DirectiveLocation::FRAGMENT_SPREAD,
                 __DirectiveLocation::INLINE_FRAGMENT,
             ],
-            args: Default::default(),
+            args: {
+                let mut args = IndexMap::new();
+                args.insert(
+                    "label",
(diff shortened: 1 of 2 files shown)


sunli829:
This is not really supporting @defer on the Fragment.

sync:
Agreed i don't know what I am doing :-)

sync:
Just checked, indeed with this the response isn't streamed back but i get a full json instead.

sync:
the end goal is to stream those fragments (which isn't working now) and to provide the label back on the subsequent chunk:

```
{"data":{"node":{"__typename":"User","id":"726995694","timeline_list_feed_units":{"edges":[{"node": 
.....
{"label":"ProfileCometTimelineFeed_user$stream$ProfileCometTimelineFeed_user_t
```

sync:
We will reopen this once when standard is properly defined. Thanks for your help

PR #153, opened 2020-06-05
add feature attribute to field attribute inside Object

bnjjj:
Code example:
```rust
#[field(feature = "users")]
pub async fn all_users(
    &self,
    context: &Context<'_>,
    after: Option<String>,
    before: Option<String>,
    first: Option<i32>,
    last: Option<i32>,
) -> FieldResult<Connection<usize, UserPosition, EmptyFields, EmptyFields>> {
    Users::fetch()
}
```

Response if you don't have feature `users` enabled:
```
{
  "errors": [
    {
      "message": "`allUsers` is only available if the features `users` are enabled",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "allUsers"
      ]
    }
  ]
}
```

### DIFF
--- a/async-graphql-derive/src/args.rs
+++ b/async-graphql-derive/src/args.rs
@@ -200,6 +200,7 @@ pub struct Field {
     pub is_ref: bool,
     pub guard: Option<TokenStream>,
     pub post_guard: Option<TokenStream>,
+    pub features: Vec<String>,
 }
 
 impl Field {
@@ -211,6 +212,7 @@ impl Field {
         let mut external = false;
         let mut provides = None;
         let mut requires = None;
+        let mut features = Vec::new();
         let mut is_ref = false;
         let mut guard = None;
         let mut post_guard = None;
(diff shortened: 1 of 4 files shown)


sunli829:
That's great! I like people who create PR directly. 😁

sunli829:
Could you please add to the Subscription? 😁

bnjjj:
Will do :)

bnjjj:
@sunli829 Do you have an example in subscription when we return an error ?

sunli829:
These two places, they're actually the same as `Object`.

https://github.com/async-graphql/async-graphql/blob/076ccf470804896da9d66e1e6dd70f55cfc76caf/async-graphql-derive/src/subscription.rs#L343

https://github.com/async-graphql/async-graphql/blob/076ccf470804896da9d66e1e6dd70f55cfc76caf/async-graphql-derive/src/subscription.rs#L242

sunli829:
Here is an example of a subscription returning an error.

https://github.com/async-graphql/async-graphql/blob/076ccf470804896da9d66e1e6dd70f55cfc76caf/tests/subscription.rs#L207

bnjjj:
@sunli829 done ! let me know if it's good for you :)

PR #155, opened 2020-06-05
Add the ability to specify full config for GraphQLPlayground

### LINKED ISSUE #139: Introduce the ability to supply a header for graphiql aka graphql playground.
Currently it's possible to hit secured graphql endpoints specifying the header value manually via browser html editor:
```
{
   "Authorization":"Bearer <token from request>"
}
``` 

It would help a lot if this could be done in a more convenient manner, like so:

```
    Ok(HttpResponse::Ok()
        .content_type("text/html; charset=utf-8")
        .body(playground_source("/graphql", None, Some(token)))
```

or even 

```
    Ok(HttpResponse::Ok()
        .content_type("text/html; charset=utf-8")
        .body(playground_source("/graphql", None, Some(header_map)))

```

which would be more generic.

ruseinov:
Resolves #139

### DIFF
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -10,7 +10,7 @@ use itertools::Itertools;
 
 pub use graphiql_source::graphiql_source;
 pub use multipart_stream::multipart_stream;
-pub use playground_source::playground_source;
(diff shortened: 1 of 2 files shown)


sunli829:
Thank you very much! 😀

PR #156, opened 2020-06-06
Update FnMut to FnOnce for query closure

bnjjj:


sunli829:
Thank you very much! 😀

PR #166, opened 2020-06-09
Update name of scalar DateTime to DateTimeUtc

bnjjj:
I'm currently migrating from Juniper and I use a tool to check if there isn't any breaking change in my schema. Do you mind if we use the same name as Juniper here https://github.com/graphql-rust/juniper/blob/2cb96d0fc4386b1ec5de8912cb0df0f271fa7cf8/juniper/src/integrations/chrono.rs#L53 for this scalar ?


sunli829:
Thank you! 😁

PR #171, opened 2020-06-11
Add allow(unused_braces, unused_parens) attribute

mwilliammyers:
This fixes false positive warnings for the stable compiler.

This can probably be removed once https://github.com/rust-lang/rust/pull/70789 lands on stable?


sunli829:
Thank you very much.

PR #175, opened 2020-06-14
Added u32 as ScalarType

MidasLamb:


sunli829:
Thank you.

mwilliammyers:
Why is `Int64` being mapped to `u32`?

sunli829:
I made some changes.

u32 -> UInt64
u64 -> UInt64

MidasLamb:
Sorry for the late reply. 
@mwilliammyers, I based it on the implementation of u64, but that needed to be corrected as wel it seems

PR #179, opened 2020-06-16
Update date time parsing to support more formats

jatsrt:
This includes the JS standard format that is not compatible with the `%+` time format.

`%+` is fairly strict in its parsing, however `DateTime<Utc>` is capable of parsing more standard formats by default. 

This allows for JSON standard formats such as:
`2020-06-23T10:00:00-09:00`


sunli829:
Thank you.😁

PR #204, opened 2020-07-07
Document default values

### LINKED ISSUE #201: Update documentation on `#[arg(default)]`
The [documentation for `#[arg(default)]`](https://docs.rs/async-graphql/1.16.5/async_graphql/attr.Object.html#field-argument-parameters) just says that it takes a string, and doesn't at all mention that it takes a literal, can call `::default()` and `default_with` - this behaviour is only documented in [the book](https://async-graphql.github.io/async-graphql/en/default_value.html).

Koxiaet:
Closes #201. I also fixed a couple grammatical errors. I haven't tested whether _all_ the changes are accurate, but I am assuming they are.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,15 +34,15 @@
 //!
 //! ## Features
 //!
-//! * Fully support async/await
+//! * Fully supports async/await
 //! * Type safety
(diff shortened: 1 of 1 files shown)


sunli829:
Thank you very much! 😁

Issue #146, opened 2020-06-04
Relay @defer requires a label in the second chunk


sync:
Trying to use defer with relay and get the following error:

```
Invariant Violation: RelayModernQueryExecutor: invalid incremental payload, expected `path` and `label` to either both be null/undefined, or `path` to be an `Array<string | number>` and `label` to be a `string`
```

Found here: https://github.com/facebook/relay/blob/2859aa8df4df7d4d6d9eef4c9dc1134286773710/packages/relay-runtime/store/RelayModernQueryExecutor.js#L1267

<img width="1122" alt="Screen Shot 2020-06-04 at 5 44 56 pm" src="https://user-images.githubusercontent.com/21725/83729673-47978a00-a68b-11ea-92a4-22ea0096dde8.png">

This is what my chunk look like:

```

---
Content-Type: application/json
Content-Length: 1113

{"data":{"room":{"title":"28 Elizabeth Street, Malvern","host":{"contactIq":null}}}}
---
Content-Type: application/json
Content-Length: 61

{"path":["room","host","contactIq"],"data":{"fullName":null}}
-----
```

Looking at what facebook.com does with their @defer they definitely have label in their response payload

```
{"data":{"node":{"__typename":"User","id":"726995694","timeline_list_feed_units":{"edges":[{"node": 
.....
{"label":"ProfileCometTimelineFeed_user$stream$ProfileCometTimelineFeed_user_timeline_list_feed_units","path":["node","timeline_list_feed_units","edges",1],"data":{"node":
```

sync:
I can see that your example works with apollo but I was super keen to get this also working with relay and it seems that there is a label missing in the second chunk. Do you have any idea of what it could be ?

sunli829:
I'll see how to RelayModernQueryExecutor compatible.

sync:
it seems that it is happy with any label, i've hacked around on the client 
```
        console.log('parts: ', parts);
        if (parts[1].path) {
          parts[1] = {
            ...parts[1],
            label: 'fake',
          };
          console.log('has path');
        }
```

and as soon as i add a label it is fine

sync:
so basically if the server would add a path from:
```
{"path":["room","host","contactIq"],"data":{"fullName":null}}
```

to

```
{"label": "SOMETHING_HERE", "path":["room","host","contactIq"],"data":{"fullName":null}}
```

it stops complaining, i don't understand how those labels are computed tough and if adding a random label has any consequences.

[...]

PR #210, opened 2020-07-13
Feature / Batch Requests

### LINKED ISSUE #198: Apollo batch request support
Hi!
Do you have any interest in supporting [batch requests](https://www.apollographql.com/blog/batching-client-graphql-queries-a685f5bcd41b)?

msdinit:
Closes #198 

My initial approach was to only expose `BatchGQLRequest` (and same for QueryBuilder / Response), but I could not find any specification for how streaming responses should be handled in batch request case (also it appears that you have plans to change streaming implementation anyway https://github.com/async-graphql/async-graphql/issues/146). It also would mean that return types would all be `BatchQueryResponse` which seemed like too big of a public API change - let me know your thoughts.

So all I did is to add batch support as an option, and add batch handlers to the integrations.


sunli829:
Thanks for hark work, I have reviewed your changes, and indeed, as you said, this is indeed a very big change.

In fact, I plan to delete streaming requests. After all, this feature has not yet entered the official specification, and the current architecture is very difficult to support these features.

So if you don't consider this function, can you extend `GQLRequest` instead of adding `BatchGQLRequest` type, so that `async_graphql_warp::graphql_batch` is no longer needed, `async_graphql_warp::graphql` can support batch requests and try to keep the API concise.

Thanks again!

msdinit:
Thanks for your kind words.

As for extending `GQLRequest` instead of adding `BatchGQLRequest` - I don't think this is possible, since we need the `BatchGQLRequest` enum to be able to deserialize both single and batch requests easily.

I can however make only `BatchGQLRequest` and `BatchQueryResponse` public and make them main input and output types of the library (meaning for example calling `execute` on schema will also return `BatchQueryResponse`, just with a `Single` variant).
This will mean updating a bunch of tests, but probably won't affect users much? They should automagically gain batch query support.

One awkward part I find with batch query support is the `QueryBuilder`, specifically support for batches - constructing it from request is nice and straightforward, but if somebody wants to construct query with a builder from the ground up, I can't find a nice way represent building either a single query or a batch query on a single struct. Although I still have an idea for a nice manual batch query-building experience I want to try out first.

So to summarize:
- Are we OK with making `Batch*` types as main exposed types? (btw, my implementation is heavily inspired by `juniper`, so I may simply be missing other possible solutions)
- If so, should I just remove all mentions of streaming support / replace with mention of it not being supported until a specification settles? (multipart requests for uploading files will still be supported both for single requests and batches)
- I'll put some more work into making better API for `BatchQueryBuilder` if it's to become the main interface - main goal is that building a query should not require juggling both `QueryBuilder` and `BatchQueryBuilder` at the same time.

In the meantime I'm converting this to a draft :)

sunli829:
Is it feasible to replace the original GQLRequest with BatchGQLRequest and try to keep the original interface?
I'm going to open a branch to remove all the code for streaming requests.

sunli829:
Or add `BatchRequest` and `BatchResponse` as you said, which I think is also a good implementation.

sunli829:
I've removed all the code for streaming requests from the `BatchRequest` branch.

sunli829:
> If so, should I just remove all mentions of streaming support / replace with mention of it not being supported until a specification settles? (multipart requests for uploading files will still be supported both for single requests and batches)
I'll put some more work into making better API for BatchQueryBuilder if it's to become the main interface - main goal is that building a query should not require juggling both QueryBuilder and BatchQueryBuilder at the same time.

I thought about it and thought it was a good idea. Let's move forward with this goal.

Thank you again for your efforts on this project. 😀

sunli829:
Keep GQLRequest, but GQLBatchRequest is an enumeration, can hold one or more GQLRequests, corresponding to GQLBatchResponse, my understanding is correct?

msdinit:
Yes

msdinit:
Sorry it's taking so long - I know you want to stabilize API :)

So this still requires a bit of polish, but the main things are implemented as we discussed above. Wanted to know your opinion before proceding to documentation/removing stream response code.
Streaming response support is removed (`execute_stream` etc), and by default all integrations are batch-enabled, without separate batch functions.
Just to test user-facing changes, I [added a PR](https://github.com/async-graphql/examples/pull/24) into `examples` that compiles - most changes are removing `defer` examples. As for other examples, changes are constrained to renaming types, so should not be a big deal.

For `QueryBuilder` I decided it was best to split it into `QueryBuilder` and `QueryDefinition` - first one is used to build queries, second can store `ctx_data` and execure itself on a given schema.
This way I was able to achieve nice-ish API for building the queries:
Example for single query:
```
let definition = QueryBuilder::new_single(&query_source)
.variables(vars)
.label(label)
.finish();
```
And it can be chained further:
```
let resp = QueryBuilder::new_single(&query_source)
.variables(vars)
.label(label)
.finish()
.execute(&schema)
.await;
```

Same for batches:
```
let resp = QueryBuilder::new_batch(&first_query_source)
.variables(vars_one)
.next(&second_query_source)
.variables(vars_two)
.finish()
.execute(&schema)
.await;
```

sunli829:
Thank you for your work, but this change is too big, I want to evaluate it. 😀

sunli829:
I deleted all the code about streaming requests.

sunli829:
I have a suggestion to rename `BatchGQLRequest` to `GQLRequest` and `GQLRequest` to `GQLRequestPart`. What do you think?

msdinit:
Sure, sounds good to me :smile:

msdinit:
Done with renames, also renamed QueryDefinition -> QueryDefinitionPart

sunli829:
API is still not easy to use, can you find ways to simplify?

msdinit:
Hi!
Terribly sorry for prolonged silence, got pulled into another project.


Regarding simplifying API - frankly, I've got no ideas. Request needs to be an ENUM to seamlessly handle both single and batch requests, and consequently response needs to be ENUM as well. I assume working with the resulting response enum is what you say isn't easy?

One change I can think of is that `QueryDefinition.execute` need not return `BatchQueryResponse` - that should eliminate at least some of the friction with this implementation. This way changes can be limited only to interactions with `QueryDefinition`, what do you think?

sunli829:
Let me think about it. In fact, I haven't thought of a good way. 😂

msdinit:
It's easier for me to think while writing code, so [here's how](https://github.com/async-graphql/async-graphql/compare/master...msdinit:feature/batch-requests-keep-schema-execute?expand=1) the changes look if `QueryDefinition.execute` returns `Result<QueryResponse>` instead of `BatchQueryResponse` :smile: 

Will see if I can come up with something else as well

sunli829:
Thank you for your efforts. I think it is now gradually moving in the right direction.
I think we should not expose the batch API to users who do not need to use them, only when they really need batch. For example, `async_graphql_actix_web::BatchGQLRequest`, this type confuses me. 😁

Thanks again, because this is a huge change, I hope to keep the api forward compatibility as much as possible.

sunli829:
Because there are many problems in the API design of the 1.x version, the code for batch queries is not easy to use. This is my problem. I apologize for that.
In version 2.0, we have redesigned a lot of APIs and added support for batch queries. Using a lot of code you have already done. Thank you very much for your contribution.
In version 2.0, you can use `http::receive_batch` to parse batch queries, and then use `Schema::execute_batch` to execute it. If you want, you can review our current implementation, your opinion is very important to us.
Thanks again! 🙂 @msdinit

Koxiaet:
@sunli829 Right now, batched operations are implemented 
 separately from everything else (e.g. `BatchRequest` and `Request`). It might be simpler to treat everything as a batch request, and have singular requests be a special case where the number of requests in the batch is 1.

sunli829:
> @sunli829 Right now, batched operations are implemented 
>  separately from everything else (e.g. `BatchRequest` and `Request`). It might be simpler to treat everything as a batch request, and have singular requests be a special case where the number of requests in the batch is 1.


Make `Request` to an internal type, `BatchRequest` instead of `Request`, right?

Koxiaet:
Actually nevermind I didn't read the code properly. It's all good, sorry for the inconvenience!

sunli829:
> Actually nevermind I didn't read the code properly. It's all good, sorry for the inconvenience!


Actually I also considering whether to use `BatchRequest` instead of a `Request`. 😃

PR #232, opened 2020-08-08
3% speedup in chat run by allocating vector up front

MidasLamb:
According to flamegraph, growing the vector (for the chat run) takes a large chunk of time: 
image

This will take an optimistic approach and assume no fields will be skipped (which will be true in the majority of cases), and initialize the array with a capacity of the amount of items, lessening the amount of resizes that need to be done.

On my pc this resulted in a 3% win for the "chat run" benchmark, but also a 2% slowdown on "simple parse" benchmark, but since "chat run" is in the order of milliseconds, and more representative of a real workload, I think this tradeoff is wel worth


sunli829:
Thank you! 😀

Issue #223, opened 2020-08-02
Clean up crate root


Koxiaet:
Currently the crate root contains a lot of items, which can be quite intimidating to new users. I propose that two new modules are created, `types` which contains commonly used GraphQL types and `protocol` which contains lower-level GraphQL types (this name is up for debate) that the user won't typically use. The following items should be moved:

```
Any -> types::Any
EmptyMutation -> types::EmptyMutation
EmptySubscription -> types::EmptySubscription
ID -> types::Id
Json -> types::Json
Lookahead -> protocol::Lookahead
Number -> types::Number
OutputJson -> types::OutputJson
Pos -> protocol::Pos
Positioned -> protocol::Positioned
QueryPathNode -> protocol::QueryPathNode
RuleError -> protocol::RuleError
Upload -> types::Upload
Variables -> protocol::Variables
WebSocketTransport -> http::WebSocketTransport
ParseRequestError -> protocol::ParseRequestError
Value -> types::Value
```

And of course the crate will also `pub use types::*` and `pub use protocol::*` so it won't be a breaking change. However, it will make it easier for people to read the library docs as the types will be better organized.

`types` could also be called `scalars` and just contain scalars.

sunli829:
> Currently the crate root contains a lot of items, which can be quite intimidating to new users. I propose that two new modules are created, `types` which contains commonly used GraphQL types and `protocol` which contains lower-level GraphQL types (this name is up for debate) that the user won't typically use. The following items should be moved:
> 
> ```
> Any -> types::Any
> EmptyMutation -> types::EmptyMutation
> EmptySubscription -> types::EmptySubscription
> ID -> types::Id
> Json -> types::Json
> Lookahead -> protocol::Lookahead
> Number -> types::Number
> OutputJson -> types::OutputJson
> Pos -> protocol::Pos
> Positioned -> protocol::Positioned
> QueryPathNode -> protocol::QueryPathNode
> RuleError -> protocol::RuleError
> Upload -> types::Upload
> Variables -> protocol::Variables
> WebSocketTransport -> http::WebSocketTransport
> ParseRequestError -> protocol::ParseRequestError
> Value -> types::Value
> ```
> 
> And of course the crate will also `pub use types::*` and `pub use protocol::*` so it won't be a breaking change. However, it will make it easier for people to read the library docs as the types will be better organized.
> 
> `types` could also be called `scalars` and just contain scalars.

It looks great! 😁

PR #252, opened 2020-08-31
Remove dependency on serde_derive, satisfy Clippy, improve docs

Koxiaet:
Since `serde` reexports `serde_derive`, I removed the dependency on it and did everything through Serde itself. I also removed usages of `extern crate` and `#[macro_use]` since they are obsoleted by the 2018 edition.

I tried to add docs where I could but I left out the places that I didn't understand well enough to properly document - it will show warnings for this.

I'll make some more docs improvements tomorrow, and eventually get round to solving #223.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ anyhow = "1.0.26"
 thiserror = "1.0.11"
 async-trait = "0.1.30"
 serde = "1.0.104"
-serde_derive = "1.0.104"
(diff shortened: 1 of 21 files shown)


Koxiaet:
So CI failed, but I don't know why since I didn't actually change the code that caused it to fail. Could this be merged anyway?

sunli829:
Leave the rest to me, thank you for your efforts. 😁

PR #253, opened 2020-09-01
Improve book and fix serde dependency

Koxiaet:
The correct feature to enable deriving for Serde is `derive`, not `serde_derive` [source](https://github.com/serde-rs/serde/blob/master/serde/Cargo.toml#L34).

I went through and fixed mistakes in the Book. I also ran a spell checker so it should all be correct.


sunli829:
Great! thank you very much!  😁

Issue #222, opened 2020-07-30
Rocket support?


numToStr:
Is [rocket](https://github.com/SergioBenitez/Rocket) support planned? As rocket can be built on stable rust.

I really love this lib and using this with rocket will be awesome.

sunli829:
In the plan, it should be added in the next few days.

sunli829:
I tried it. `Rocket` does not currently support stable, so this work needs to be postponed. If you are familiar with `Rocket`, would you like to contribute an `async-graphql-rocket`? 😁

numToStr:
I am just getting started with rust and I was trying out rocket. Maybe I'll just wait for the `Rocket` support. 🤘🏻

nicolaiunrein:
Isn’t Rocket a synchronous framework?

numToStr:
I read that Rocket 0.5 will be async

Weasy666:
I'm currently trying to implement this [here](https://github.com/simplificAR/async-graphql/tree/master/async-graphql-rocket), but...i hit this [problem](https://gist.github.com/Weasy666/2a230114b74aed7a06a7cb90cab6aeb3).
I think the problem could be the combination of generics and the Rockets route macros, but i'm not sure. Need to investigate some more, but wanted to tell that i'm working on this.

sunli829:
> I'm currently trying to implement this [here](https://github.com/simplificAR/async-graphql/tree/master/async-graphql-rocket), but...i hit this [problem](https://gist.github.com/Weasy666/2a230114b74aed7a06a7cb90cab6aeb3).
> I think the problem could be the combination of generics and the Rockets route macros, but i'm not sure. Need to investigate some more, but wanted to tell that i'm working on this.

Is it possible to avoid using generics by defining new types?
```rust
type MySchema = Schema<.....>;
```

Weasy666:
Don't we need generics, for people to be able to provide their own schema?

sunli829:
> Don't we need generics, for people to be able to provide their own schema?

Sorry, I don’t know how to use rocket. 😂
This seems to be a problem with Rocket.

Weasy666:
Ok...it is compiling, see [custom_handler](https://github.com/simplificAR/async-graphql/tree/custom_handler/async-graphql-rocket) branch on my fork.

[...]

PR #256, opened 2020-09-04
Add Rocket integration

Weasy666:
This PR addresses #222 and adds initial Rocket.rs support to `async-graphql`. Another [PR](https://github.com/async-graphql/examples/pull/26) adds the `StarWars` example to the `examples` repo.

Initial support means, that it currently depends on a specific commit of Rocket's github repository, which needs to be changed as soon as version 0.5 gets released. It also means, that subscription is not support as that need Websocket support (this is correct, right?) and that is currently not supported or to be more precise, integrated into Rocket.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ members = [
     "async-graphql-actix-web",
     "async-graphql-warp",
     "async-graphql-tide",
-    #    "async-graphql-lambda",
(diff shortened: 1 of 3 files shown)


sunli829:
Great, thank you very much. After the `rocket 0.5` is released, we may publish it to crates.io with only a few minor modifications.

PR #257, opened 2020-09-06
Rewrite async-graphql-parser

Koxiaet:
This is a large one.
- I rewrote most of `async-graphql-parser`.
    - I removed the `schema` module entirely, since it wasn't used anywhere. This means `async-graphql-parser` can no longer be used to parse non-executable documents (such as schema definition documents), but that's not a particularly useful thing to do anyway.
    - Since there's nothing to contrast `schema` with, the `query` module has been renamed to just `types`.
    - I added docs and references to the spec throughout.
    - I replaced the mutable-variable-and-loop parsing approach used before with a much more strict `next()`-based sequential approach, which has less mutable variables and makes the intent more clear.
    - I moved some invariants to the grammar itself, allowing the parsing functions to be infallible.
    - I changed `OperationDefinition` to be a struct with a `ty` field (`query`/`mutation`/`subscription`) instead of an enum with repeated fields. This reduces duplicated code in many places.
    - The single anonymous `SelectionSet` special case gets translated to a regular `Query` at parse time. This reduces the number of special cases necessary.
    - Instead of having one `Document` type that can either have a list of definitions or the current operation and fragment map, I created two separate types `Document` and `ExecutableDocument` for the former and latter uses respectively. This helps distinguish the two types.
    - Big one; I removed the `Deref` impl for `Positioned`, since it isn't a smart pointer or simple owned container (relevant: [this URLO thread](https://users.rust-lang.org/t/understanding-the-perils-of-deref/47958)).
        - This means that most of the changes are just adding a bunch of `.node`s, and uh well...
```sh
$ rg -o '\.node' | wc -l
407
```
🙃
- I removed `Deref` for `Variables` and made it a completely transparent newtype around `BTreeMap<String, Value>`. The ergonomics of a public newtype should hopefully make up for the loss of ergonomics of `Deref`.
- I also refactored various random parts around the code base, so there are some smaller changes not listed here.
- See `src/context.rs` line 35; I didn't change the behaviour of the code but I noticed there might be a bug.

Edit: Oh no, CI fails because it uses an outdated version of Rust. What should I do?


sunli829:
👏 Thank you for your contribution! You removed the schema parsing module. Although `async-graphql` does not use it, it is used by `codegen-for-async-graphql-renderer`. So, can this module be restored? 🙂

sunli829:
> See src/context.rs line 35; I didn't change the behaviour of the code but I noticed there might be a bug.

I changed this to this, which I think is easier to use. 🙂

```rust
    /// Parse variables from JSON object.
    /// If this `value` is not a `JsonObject`, then an empty `Variables` instance will be returned.
    pub fn parse_from_json(value: serde_json::Value) -> Self {
        if let Value::Object(obj) = value.into() {
            Self(obj)
        } else {
            Default::default()
        }
    }

```

Koxiaet:
> Although async-graphql does not use it, it is used by codegen-for-async-graphql-renderer. So, can this module be restored? 🙂

Sorry, I did not realize that! I will restore it.

Koxiaet:
I have restored the module, you can see the commit notes for more. There are some things I forgot to mention there so I'll mention them here:

- I removed the infallible conversions from GraphQL values to JSON values, and replaced them with fallible conversions (since not all JSON keys are GraphQL names and not all GraphQL values (e.g. Upload & Variable) are valid JSON values).
- I renamed `Document`/`ExecutableDocument` to `ExecutableDocument`/`ExecutableDocumentData` since `Document` is no longer unique (see `ServiceDocument`).

There's quite a bit of code duplication to implement the const/non-const counterparts of values, directives, etc. It's a bit of a shame but I don't see an easy way around it other than macros (which is probably overkill).

Also, I'm not complete sure about the `Name` newtype. Having more type-safety is nice, but I don't know if it's necessary.

Koxiaet:
Ok, so I finally fixed all the tests (hopefully). The problem was that `Upload` validated only when the input is a string that begins with `file:`, and changing it to validate on regular `Upload`s made it work. Why was is like that, and why did it work before?

sunli829:
> I have restored the module, you can see the commit notes for more. There are some things I forgot to mention there so I'll mention them here:
> 
> * I removed the infallible conversions from GraphQL values to JSON values, and replaced them with fallible conversions (since not all JSON keys are GraphQL names and not all GraphQL values (e.g. Upload & Variable) are valid JSON values).
> * I renamed `Document`/`ExecutableDocument` to `ExecutableDocument`/`ExecutableDocumentData` since `Document` is no longer unique (see `ServiceDocument`).
> 
> There's quite a bit of code duplication to implement the const/non-const counterparts of values, directives, etc. It's a bit of a shame but I don't see an easy way around it other than macros (which is probably overkill).
> 
> Also, I'm not complete sure about the `Name` newtype. Having more type-safety is nice, but I don't know if it's necessary.

Thank you, async-graphql is getting better and better! 😁

sunli829:
> Ok, so I finally fixed all the tests (hopefully). The problem was that `Upload` validated only when the input is a string that begins with `file:`, and changing it to validate on regular `Upload`s made it work. Why was is like that, and why did it work before?

Haha, this is a bug. A long time ago Upload was represented by a string prefixed with `file:`.

PR #260, opened 2020-09-09
add context in ResolveInfo and add headerMap in warp integration

bnjjj:
The use case here is to let the user indicate in a header if we want to enable or disable a specific feature in an extension. So basically I added support for headers in warp integration and added context inside `ResolveInfo` to let me pass header data to my extension for a query inside the context.

### DIFF
--- a/src/extensions/mod.rs
+++ b/src/extensions/mod.rs
@@ -4,7 +4,7 @@ mod apollo_tracing;
 mod logger;
 mod tracing;
 
-use crate::context::{QueryPathNode, ResolveId};
+use crate::context::{Context, QueryPathNode, ResolveId};
 use crate::{Result, Variables};
 
 pub use self::apollo_tracing::ApolloTracing;
@@ -27,6 +27,9 @@ pub struct ResolveInfo<'a> {
 
     /// Current path node, You can go through the entire path.
     pub path_node: &'a QueryPathNode<'a>,
+
+    /// Context
+    pub context: &'a Context<'a>,
 
     /// Parent type
     pub parent_type: &'a str,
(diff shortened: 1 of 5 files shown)


sunli829:
Adding the output of the `HeaderMap` to the output of the `graphql` filter is actually unnecessary because you can combine it yourself where you can call it.

src/extensions/mod.rs
```diff
@@ -56,7 +59,7 @@ pub trait Extension: Sync + Send + 'static {
     fn validation_end(&mut self) {}
 
     /// Called at the begin of the execution.
-    fn execution_start(&mut self) {}
+    fn execution_start<'a>(&mut self) {}
```
sunli829, line 62: It looks quite strange here. 🙂

sunli829:
I made changes to ResolveInfo in the master branch.

sunli829:
Get the `HeaderMap` in the request, I believe this is the correct usage of warp. 🙂

```rust
let graphql_post = async_graphql_warp::graphql(schema)
        .and(warp::header::headers_cloned())
        .and_then(|(schema, builder): (_, QueryBuilder), headers| async move {
            let resp = builder.execute(&schema).await;
            Ok::<_, Infallible>(GQLResponse::from(resp))
        });
```

src/extensions/mod.rs
bnjjj: Oups it was from a previous implementation, need to delete

bnjjj:
great thanks for adding the context in `ResolveInfo` on master. We can close this pull-request now :)

sunli829:
I deleted the `context` field in `ResolveInfo`, but added the `data_XXX` methods.

PR #269, opened 2020-09-19
Add more trait implementations to MaybeUndefined

### LINKED ISSUE #267: Derive additional implementations for MaybeUndefined
It would be very helpful if `MaybeUndefined<T>` implemented:
- `Debug`
- `Clone`
- `Serialize`
- `Deserialize`

Iff `T` implements those, so we don't require `T` to implement them.

Is everyone ok with that?

I think that is a conservative enough list?

mwilliammyers:
Closes #267


sunli829:
thank you very much!😃

PR #275, opened 2020-09-23
Export description with `Registry::export_sdl`

Weasy666:
`Registry::export` now also export the description/docs of all types when called with `federation: false`. `Schema::sdl` always exports the description.


sunli829:
Thank you very much!

PR #284, opened 2020-09-27
Add description to SimpleObject

esemeniuc:


docs/en/src/define_simple_object.md
```diff
@@ -8,6 +8,7 @@ The example below defines an object `MyObject` which includes the fields `a` and
 use async_graphql::*;
 
 #[derive(SimpleObject)]
+#[graphql(desc = "An optional description")]
```
sunli829, line 11: I recommend this way more. 😁

```rust
/// An optional description
#[derive(SimpleObject)]
struct MyObject {
}
```

esemeniuc: Unfortunately it doesn't show up in the schema or the graphql playground

sunli829: > Unfortunately it doesn't show up in the schema or the graphql playground. This is handy for clients and consumers of the graphql service

In fact it can. 😁

esemeniuc: Ah I see, requires a bit of looking deeply. Can we add something to the docs to make this more clear?

sunli829: The `desc` attribute will soon be removed and will only support rustdoc in the future.😃

sunli829: Perhaps you could help us add a chapter on how to add a description?😊

PR #291, opened 2020-09-30
Rework errors

Koxiaet:
See the commit message for the changes. Still need to merge master (again -_-) but other than that it's complete.


sunli829:
I am merging code.

sunli829:
I restored the `MergedSubscription` that you deleted by mistake. 😂

sunli829:
Thanks for the improvement, now the error type looks much better. 🙂

sunli829:
If there are no other questions, can this be merged?

Koxiaet:
Oh, thanks!

> I restored the MergedSubscription that you deleted by mistake. 😂

That wasn't actually by mistake; the implementation wasn't used. Basically the problem was that the subscription trait requires that the stream does not outlive `self` (as we want the stream to be able to borrow from `self`). Before, this was no problem for the derive implementation of merged objects: I just created a self-referential stream with the `async_stream` macro, roughly:

```rust
async_stream::stream! {
    let obj = create_merged_object;
    let mut stream = obj.create_stream();

    while let Some(item) = stream.next() {
        yield item;
    }
}
```

However since `create_stream` now returns an `Option`, we run into a problem:

```rust
let obj = create_merged_object;
let stream = obj.create_stream();

match stream {
    // Returning the stream directly doesn't work as then the stream will
    // outlive `obj` which is destroyed at the end of this function.
    // We also can't use the async stream macro as `obj` can't be moved
    // if `stream` references it.
    Some(stream) => /* ??? */,
    None => None,
}
```

The workaround I came up with was to ditch the subscription implementation of merged objects entirely and move the logic into the derive macro:

```rust
None
    .or_else(|| self.0.create_field_stream())
    .or_else(|| self.1.create_field_stream())
    .or_else(|| self.2.create_field_stream())
    // etc
```

That was the hack I mentioned in the commit message.

To solve this problem in a better way we could either also move the logic for resolving objects into the derive macro, or another idea I had which is to have the object trait have a pure function that returns a `&'static [(&'static str, ResolverFn)]` or something similar.

sunli829:
> If there are no other questions, can this be merged?

You are right, I didn't think about this. 😁👍🏻

I decided to delete the `MergedObject` and `MergedSubscription` macros.

`MergedObject` can be replaced by the `flatten` property.

```
#[derive(Object)]
struct MyObj {
     #[graphql(flatten)]
     obj: InnerObj,
}
```

But I have not thought of a better syntax for subscription.

sunli829:
It doesn't work, `Interface` is implemented through function forwarding, so the `Object` created in this way cannot support `Interface`.

sunli829:
Wow, I did not notice that you have actually modified the implementation of the `MergedSubscription` macro! Thank you so much! ❤️

Koxiaet:
Oh, that's a good point. Maybe we should reintroduce the old system then.

Issue #293, opened 2020-10-02
Rework guard parsing


AurelienFT:
Hello,

I use the repository for a project in my company and I want to help a bit to contribute to the amazing work you guys doing.

I figured out that guard only support (implicitly) `and` operator which accept `x` numbers of rules and the implementation is very specific to this operator.

If we take [graphql-shield](https://github.com/maticzav/graphql-shield) (a huge guard module for graphql api in typescript) as a example they implement some others operators : 

- and
- or
- not 
- chain 
- race

I would like to rework the actual implementation to make a recursive parsing on operators and so be able to parse things like : 
```
#[graphql(guard(
     and(
         or(
             RoleGuard(role = "Role::Admin"),
             RoleGuard(role = "Role::Guest")
         ),
         UserGuard(username = r#""test""#)
     )
))]
```

You can find explanations about each operators [here](https://github.com/maticzav/graphql-shield#logic-rules).

It's my first work on this project and on macros in general so I started look deep into `generate_guards()` and I understand how the function work. I start working on a first version of the implementation and I will submit a PR in draft soon.

sunli829:
Great, look forward to it! Post guard I think you can delete it, it's not too useful. Attributes parsing for procedural macros is recommended use `darling` crate, which is very handy. 😃

AurelienFT:
I don't know `darling`, I started doing the parsing the same functions that you used. Maybe I can make a v2 with `darling` after I practice with macros

AurelienFT:
I don't think implement the `not` operator seems relevant because we can't return the error defined in the `check` function of the user.

What do you think about it @sunli829 ?

sunli829:
> I don't think implement the `not` operator seems relevant because we can't return the error defined in the `check` function of the user.
> 
> What do you think about it @sunli829 ?

you're right. 😁

AurelienFT:
Implemented with https://github.com/async-graphql/async-graphql/pull/296

PR #296, opened 2020-10-03
Rework guard

AurelienFT:
Refer to #293


AurelienFT:
@sunli829 If you have some times I would like to have a little bit of help. I want to implement the chain operator which work like the guard system currently implemented on master. But I want to implement it with a vec of rules not a chain of and operator like now.

I know how to it in the parsing system (in derive subfolder) but I don't understand how to implement the operator in `src/guard.rs` which take `Vec<Guard>` has parameter.

If you can provide me some help it will be really cool !

sunli829:
Hmm, is it possible to implement `Guard` for `Vec<T:Guard>`?
But it would be better if you could not use `Vec`? 😂

AurelienFT:
Do you think it's better to chain the `and` operator with two rules than doing a Vector of rule ?

sunli829:
> Do you think it's better to chain the `and` operator with two rules than doing a Vector of rule ?

Yes, I think this can avoid additional memory allocation and dynamic dispatch. 😁

AurelienFT:
Ok I will implement it that way so. Thanks you

sunli829:
> Ok I will implement it that way so. Thanks you

Because `guard` is created every time the `resolver` function is called, right?

AurelienFT:
> > Ok I will implement it that way so. Thanks you
> 
> Because `guard` is created every time the `resolver` function is called, right?

`Guard` macro is expended on the code of the `resolve_field` function so, if I understand well, it's created one time at the creation of the field resolver. Like this (example of expanded `resolve_field`) : 
```
impl async_graphql::resolver_utils::ContainerType for Query {
                fn resolve_field<'life0, 'life1, 'life2, 'async_trait>(
                    &'life0 self,
                    ctx: &'life1 async_graphql::Context<'life2>,
                ) -> ::core::pin::Pin<
                    Box<
                        dyn ::core::future::Future<
                                Output = async_graphql::ServerResult<
                                    ::std::option::Option<async_graphql::serde_json::Value>,
                                >,
                            > + ::core::marker::Send
                            + 'async_trait,
                    >,
                >
                where
                    'life0: 'async_trait,
                    'life1: 'async_trait,
                    'life2: 'async_trait,
                    Self: 'async_trait,
                {
                    #[allow(
                        unused_parens,
                        clippy::missing_docs_in_private_items,
                        clippy::needless_lifetimes,
                        clippy::ptr_arg,
                        clippy::trivially_copy_pass_by_ref,
                        clippy::type_repetition_in_bounds,
                        clippy::used_underscore_binding
                    )]
                    async fn __resolve_field(
                        _self: &Query,
                        ctx: &async_graphql::Context<'_>,
                    ) -> async_graphql::ServerResult<
                        ::std::option::Option<async_graphql::serde_json::Value>,
                    > {
                        if ctx.item.node.name.node == "value" {
                            async_graphql::guard::GuardExt::and(
                                RoleGuard {
                                    role: (Role::Admin).into(),
                                },
                                UserGuard {
                                    username: ("test").into(),
                                },
                            )
                            .check(ctx)
                            .await
                            .map_err(|err| err.into_server_error().at(ctx.item.pos))?;
                            let res = _self
                                .value(ctx)
                                .await
                                .map_err(|err| err.into_server_error().at(ctx.item.pos))?;
```

sunli829:
Yes, but this creation code runs every time it resolves the field.

发自我的iPhone

AurelienFT:
> Yes, but this creation code runs every time it resolves the field.

Yes it runs at each call of the resolver. You think vector gonna need too much ressources ?

sunli829:
> > Yes, but this creation code runs every time it resolves the field.
> 
> Yes it runs at each call of the resolver. You think vector gonna need too much ressources ?

Not much, but there will always be some loss. 😁

AurelienFT:
Ready, reviews are welcome

sunli829:
Thank you very much!
Could you please add a chapter `guard.md` to the `book`? 😀

AurelienFT:
Done.

sunli829:
Thank you so much! 😃

PR #304, opened 2020-10-08
Derive Error with Eq

AurelienFT:
I wanted to unit test the code logic of my resolvers. But my resolvers return a `async_graphql::Error` and I can't use assert_eq! because `async_graphql::Error` do not implement equal.

I can serialize to json each error to test but I don't know if it's the best way to do it so I prefer propose this MR to add `Eq`.

Feedback are welcome.


sunli829:
Thank you! 😁

PR #321, opened 2020-10-23
Fixed redundant_closure_for_method_calls

olexiyb:
https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure_for_method_calls

The size of library before:

```
$ cargo build --release && ls -al target/release/libasync_graphql.rlib
    Finished release [optimized] target(s) in 0.07s
-rw-rw-r-- 2 olexiyb olexiyb 7847678 Oct 23 10:57 target/release/libasync_graphql.rlib
```

after

```
$ cargo build --release && ls -al target/release/libasync_graphql.rlib
   Compiling async-graphql v2.0.8 (/home/olexiyb/b100pro/async-graphql)
    Finished release [optimized] target(s) in 10.09s
-rw-rw-r-- 2 olexiyb olexiyb 7837136 Oct 23 11:05 target/release/libasync_graphql.rlib
```

around 10kB win and I think this will improve performance as well

### DIFF
--- a/src/http/multipart.rs
+++ b/src/http/multipart.rs
@@ -85,8 +85,7 @@ pub(super) async fn receive_batch_multipart(
             _ => {
                 if let Some(name) = field.name().map(ToString::to_string) {
                     if let Some(filename) = field.file_name().map(ToString::to_string) {
-                        let content_type =
-                            field.content_type().map(std::string::ToString::to_string);
+                        let content_type = field.content_type().map(ToString::to_string);
                         let mut file = tempfile::tempfile().map_err(ParseRequestError::Io)?;
                         while let Some(chunk) = field.chunk().await.unwrap() {
                             file.write(&chunk).map_err(ParseRequestError::Io)?;
--- a/src/http/multipart.rs
+++ b/src/http/multipart.rs
@@ -85,7 +85,8 @@ pub(super) async fn receive_batch_multipart(
             _ => {
                 if let Some(name) = field.name().map(ToString::to_string) {
                     if let Some(filename) = field.file_name().map(ToString::to_string) {
-                        let content_type = field.content_type().map(|mime| mime.to_string());
+                        let content_type =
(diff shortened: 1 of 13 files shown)


src/validation/rules/no_undefined_variables.rs
```diff
@@ -80,7 +80,7 @@ impl<'a> Visitor<'a> for NoUndefinedVariables<'a> {
         name: Option<&'a Name>,
         operation_definition: &'a Positioned<OperationDefinition>,
     ) {
-        let name = name.map(|name| name.as_str());
+        let name = name.map(async_graphql_value::Name::as_str);
```
Koxiaet, line 83: `Name` is in scope here, no need to write the full path.

src/validation/utils.rs
```diff
@@ -124,7 +124,7 @@ pub fn is_valid_input_value(
                     ConstValue::Object(values) => {
                         let mut input_names = values
                             .keys()
-                            .map(|name| name.as_ref())
+                            .map(std::convert::AsRef::as_ref)
```
Koxiaet, line 127: `AsRef` is in the prelude, there's no need for `std::convert::`.

src/http/multipart.rs
```diff
@@ -85,7 +85,8 @@ pub(super) async fn receive_batch_multipart(
             _ => {
                 if let Some(name) = field.name().map(ToString::to_string) {
                     if let Some(filename) = field.file_name().map(ToString::to_string) {
-                        let content_type = field.content_type().map(|mime| mime.to_string());
+                        let content_type =
+                            field.content_type().map(std::string::ToString::to_string);
```
Koxiaet, line 89: `ToString` is in the prelude, so don't have `std::string::`; this applies to all the other uses of that as well.

Koxiaet (commented):
Thanks for this! Just make sure that you're not using any qualified paths when shortened ones will do.

Koxiaet:
Seems good to merge.

PR #324, opened 2020-10-26
Modify interface SDL and bump Rocket

Weasy666:
This PR includes:
- Bump Rocket to after config revamp and change examples acordingly
- Add route ranks to Rocket to remove collision warning on startup
- Add some convenience wrappers to Rocket integration
- Change implements Interface SDL generation to adhere to [spec](http://spec.graphql.org/June2018/#sec-Interfaces). 
  Implementation of a single interface should yield a type like this `type Person implements NamedEntity`, without a `&`.

### DIFF
--- a/integrations/rocket/src/lib.rs
+++ b/integrations/rocket/src/lib.rs
@@ -33,7 +33,7 @@ mod query_deserializer;
 /// # Examples
 ///
 /// ```ignore
-/// #[rocket::post("/graphql", data = "<request>", format = "application/json")]
+/// #[rocket::post("/graphql", data = "<request>", format = "application/json", rank = 1)]
 /// async fn graphql_request(schema: State<'_, ExampleSchema>, request: BatchRequest) -> Response {
 ///     request.execute(&schema).await
 /// }
@@ -93,12 +93,12 @@ impl FromData for BatchRequest {
 /// # Examples
 ///
 /// ```ignore
-/// #[rocket::post("/graphql?<query..>")]
+/// #[rocket::post("/graphql?<query..>", rank = 2)]
 /// async fn graphql_query(schema: State<'_, ExampleSchema>, query: Request) -> Result<Response, Status> {
 ///     query.execute(&schema).await
 /// }
 ///
-/// #[rocket::post("/graphql", data = "<request>", format = "application/json")]
+/// #[rocket::post("/graphql", data = "<request>", format = "application/json", rank = 1)]
 /// async fn graphql_request(schema: State<'_, ExampleSchema>, request: Request) -> Result<Response, Status> {
 ///     request.execute(&schema).await
 /// }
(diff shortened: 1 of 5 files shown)


integrations/rocket/src/lib.rs
```diff
@@ -119,6 +119,32 @@ impl Request {
     {
         Response(schema.execute(self.0).await.into())
     }
+
+    /// Specify the operation name of the request.
+    pub fn operation_name<T: Into<String>>(self, name: T) -> Self {
```
Koxiaet, line 124: What's the use for this? `operation_name` is set by the client when the request is deserialized, servers shouldn't have to change it.

Also, is there really a benefit to being able to write `.operation_name(...).0` over `.0.operation_name(...)`? You'll have to convert it to an `async_graphql::Request` anyway, so we might as well attempt to keep the API surface small.

src/registry/export_sdl.rs
```diff
@@ -131,10 +131,7 @@ impl Registry {
                 write!(sdl, "type {} ", name).ok();
                 if let Some(implements) = self.implements.get(name) {
                     if !implements.is_empty() {
-                        write!(sdl, "implements ").ok();
-                        for interface in implements {
-                            write!(sdl, "& {} ", interface).ok();
-                        }
+                        write!(sdl, "implements {} ", implements.iter().map(AsRef::as_ref).collect::<Vec<&str>>().join(" & ")).ok();
```
Koxiaet, line 134: By the way, the previous version [adhered to the spec fine](https://spec.graphql.org/June2018/#ImplementsInterfaces). Although this version does look prettier.

integrations/rocket/src/lib.rs
Weasy666: Yeah ok, will remove it

src/registry/export_sdl.rs
Weasy666: Oh...funny that they have it different in two places.

integrations/rocket/src/lib.rs
Koxiaet: Probably all the convenience wrappers since the same logic can be applied to them all.

Weasy666: Ok, dropped commit 97b8ac3 entirely.

Weasy666: Damn...i think leaving the `data()` function would have been nice. It is much nicer to write
```rust
#[rocket::post("/?<query..>", rank = 2)]
async fn graphql_query(schema: State<'_, StarWarsSchema>, db: Db, query: Request) -> Response {
    query.data(db).execute(&schema).await
}
```
instead of 😅
```rust
#[rocket::post("/?<query..>", rank = 2)]
async fn graphql_query(schema: State<'_, StarWarsSchema>, db: Db, mut query: Request) -> Response {
    query.0.data.insert(db);
    query.execute(&schema).await
}
```

Weasy666: @Koxiaet What do you think about that? ☝

PR #331, opened 2020-10-31
Validate the query hash before persisting queries

khalsah:
Updates the `ApolloPersistedQueries` extension to verify the SHA-256 hash of queries before storing them in the cache.

1. This prevents malicious or broken clients from corrupting the cache.
2. This also more closely matches the [implementation in apollo-server](https://github.com/apollographql/apollo-server/blob/1af2792ed9e18f2bf218a29c2f4f128b6588f9ca/packages/apollo-server-core/src/requestPipeline.ts#L201-L210)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ default = [
 	"uuid",
 ]
 apollo_tracing = ["chrono"]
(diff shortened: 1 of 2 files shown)


sunli829:
Thank you very much! 😁

PR #350, opened 2020-12-02
Detect subprotocol for GraphQL Over WebSocket subscriptions

pfernie:
This PR is a WIP/request for comment. 

On 2020-09-14 a revised version of the [GraphQL over WebSocket](https://the-guild.dev/blog/graphql-over-websockets) spec was released as [graphql-ws](https://github.com/enisdenjo/graphql-ws), which differs from the de facto standard implemented by [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws).

I use this locally to enable working with the [graphql-ws](https://github.com/enisdenjo/graphql-ws) library. However:
* `graphql-playground` uses the `subscriptions-transport-ws` spec, although there is an [issue](https://github.com/graphql/graphql-playground/issues/1291) open to support the new `graphql-ws`
* The prior protocol contained a `client -> server` `GQL_CONNECTION_TERMINATE` message, whereas the new protocol does not (`Complete` in the new spec is the same as `Stop` in the old). While this change does not remove the `ClientMessage::ConnectionTerminate` variant, presumably this path is not taken under the new protocol.
* By utilizing a feature, a decision must be made a build time as to which protocol to support. A better approach may be to make this an option when instantiating the GQL schema at runtime. If it were possible to implement a "subprotocol negotiation", this would allow clients to specify which protocol they are using and both types of client could be supported, but I do not believe this can be done.

This adds a new `cargo` (non-default) feature, `graphql_ws`, which when enabled utilizes the new protocol spec.

Original [subscriptions-transport-ws protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md)
Revised [graphql-ws protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md)


sunli829:
Thank you very much for the PR, but I think it is feasible to judge the protocol used by the client through `Sec-WebSocket-Protocol`. For this reason, we may need to pass this information to `async_graphql::HTTP::WebSocket` for different implementations.
What do you think? 🙂

pfernie:
You are correct; `warp` does not expose this header via its API, so I had missed this detail. I have opened an [issue](https://github.com/seanmonstar/warp/issues/760) there to see about making this information available. When there is feedback on that issue, the behavior here could adapt to the requested protocol from the client.

pfernie:
@seanmonstar pointed out that the header is indeed accessible when the connection is initiated; the updated branch now instead implements detecting the protocol in use. It introduces a new `enum http::websocket::Protocols` which tracks this, and adds it as a field of `WebSocket`.

The actual logic largely overlaps; there is an extra variant for `ServerMessage` to support the `Next` message instead of the old spec's `Data`. Otherwise, for the `ClientMessage` enum, `serde`s `alias` directive is used to allow consumption of both protocols.

If the initiating client does not specify the `Sec-WebSocket-Protocol` header, we assume the old protocol, `Protocols::SubscriptionsTransportWS`.

pfernie:
This change breaks the integration for `actix-web`; I'm not familiar with that framework at all. Looking at the example and the documentation for `actix-web`, however, it doesn't seem to be straightforward to recover the protocol string once the `WSSubscription` has `started()`?

sunli829:
I'm trying, a breaking change may be needed here.

sunli829:
I added this feature to actix-web integration. 😁

sunli829:
@pfernie Please see what needs to be improved, thanks again! 😁

pfernie:
My use cases work with clients using either protocol, so this looks good for merge to me. Thanks for the help, and adding the test cases!

pfernie:
A small change; the [WebSocket RFC](https://tools.ietf.org/html/rfc6455#page-6) allows for the client to specify multiple supported protocols in the negotiation; this change updates the `warp` and `actix-web` integrations accordingly.

PR #379, opened 2021-01-06
Include full link

ejez:
Include the full link to ensure it is always accessible, whether from
github, doc.rs or elsewhere.


ejez:
Clicking the `Feature comparison` link from https://docs.rs/async-graphql yields an invalid link. This PR fixes the issue.

sunli829:
Thanks.👏

Issue #370, opened 2021-01-01
Setting additional HTTP properties on responses (ie. cookies & headers)


naamancurtis:
## Description of the feature

Is there any interest (or current ability to) support setting cookies/headers on the response for a query or mutation?

The example I'm thinking of would be access/refresh tokens that would be stored in `HttpOnly` cookies as a result of hitting a `refresh` mutation. I must admit I'm still relatively new to GraphQL, so the answer might just be to remove auth related functionality from GraphQL and do it via some other API technology, but I figured I'd ask.

Looking at the code, I imagine `async_graphql::Response` would have to be extended to support this?

sunli829:
> ## Description of the feature
> Is there any interest (or current ability to) support setting cookies/headers on the response for a query or mutation?
> 
> The example I'm thinking of would be access/refresh tokens that would be stored in `HttpOnly` cookies as a result of hitting a `refresh` mutation. I must admit I'm still relatively new to GraphQL, so the answer might just be to remove auth related functionality from GraphQL and do it via some other API technology, but I figured I'd ask.
> 
> Looking at the code, I imagine `async_graphql::Response` would have to be extended to support this?

I recommend sharing `Arc<Mutex<Option<String>>>` through `Context`.

Following is the example:

```rust
let cookie: Arc<Mutex<Option<String>>> = Default::default();

schema.execute(request.data(cookie)).await;

#[Object]
implement Mutation {
    async refresh_token(&self, ctx: &Context<'_>, ...) -> bool {
        *ctx::data_unchecked::<Arc<Mutex<Option<String>>>>().lock().await = Some(new_refresh_token);
        true
    }
}
```

scrogson:
I believe the question is regarding setting cookies on the HTTP response.

naamancurtis:
Yep, setting cookies/headers on the HTTP response ideally covering both use cases 
1. Setting brand new cookies/headers that don't currently exist on the request.
2. Updating existing cookies/headers with new values.

I guess the challenge here is each integration (ie. actix-web, warp etc) has its own notion of cookies & headers _(I think?)_ and the conversion from 'async_graphql::Response' to 'framework_library::HttpResponse' happens within the integration library if I understand the flow correctly?

sunli829:
This example is based on `actix-web`, and explains more fully how to get the response HTTP header from the `GraphQL` query and return it to the client.

```rust

struct;

#[Object]
impl Mutation {
[...]

PR #383, opened 2021-01-10
Update Headers to use `http::HeaderMap`

naamancurtis:
Relates to issue #370;

I think the current implementation of headers `HashMap<String, String>` doesn't fully support all possible use cases for headers that can support multiple values. A good example of this would be `Set-Cookies` where you would possibly want to set multiple Cookies on a single response.

This PR updates the underlying implementation to use `http::HeaderMap`, which was already an implicit dependency in the crate, but is now an explicit one. It's also the header implementation most (but not all) of the integration frameworks use for headers.

I've only implemented a subset of the `HeaderMap` functionality through the context, however I'm happy to go back and implement more/change things if you'd prefer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ serde_json = "1.0.48"
 spin = "0.7.0"
 thiserror = "1.0.21"
 static_assertions = "1.1.0"
+http = "0.2.3"
 
 # Feature optional dependencies
 bson = { version = "1.0.0", optional = true }
--- a/integrations/rocket/src/lib.rs
+++ b/integrations/rocket/src/lib.rs
@@ -169,17 +169,17 @@ impl<'r> Responder<'r, 'static> for Response {
         let body = serde_json::to_string(&self.0).unwrap();
 
         let mut response = rocket::Response::new();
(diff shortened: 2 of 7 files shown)


sunli829:
Haha, LGTM, thank you very much for your contribution!
Please feel free to do what you think is needed, and let me know if you encounter problems.
When you think everything is done, I will merge this PR.
Thanks again! 🙂

naamancurtis:
I think it looks good to me, I've done some playing around and it seems to add all the headers as you would expect it to.

My only slight concern is the current set up naively adds headers, either inserting them if the `key` doesn't exist, or appending them if they do. 

This could potentially lead to multiple headers for headers that are single valued, although I guess it would be up to the underlying server implementation to handle this appropriately? Any thoughts?

sunli829:
> My only slight concern is the current set up naively adds headers, either inserting them if the key doesn't exist, or appending them the key does exist.

Let me think about whether it is possible to solve this problem. 🙂

sunli829:
I think it's fine now, it's not a problem, so I'm merging this PR now.
Thanks again! 😄

PR #384, opened 2021-01-10
update response to correctly carry headers through

naamancurtis:
Apologies, made a few assumptions on my previous PR #383 which didn't hold true. I've been through and tested this more rigurously with the code from a full application and believe it works as intended.

Essentially the `http_headers` weren't being transfered across from the `query_env` to the `response` when the response object was being created.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -53,6 +53,15 @@ impl Response {
         Self { extensions, ..self }
     }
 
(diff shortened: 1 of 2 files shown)


sunli829:
Sorry, I forgot to set HTTP headers into response. 😂

PR #392, opened 2021-01-15
Add minimal-versions to Github actions

chipsenkbeil:
This is the start of adding a new Github action to verify that async-graphql properly supports the minimum versions of the dependencies listed.

For details on what this is about, check out:
- https://github.com/rust-lang/cargo/issues/5657
- https://github.com/rust-lang/cargo/issues/5656
- https://blog.illicitonion.com/rust-minimum-versions-semver-is-a-lie/

It's being encouraged that crates start to do this where possible. You can see that [tokio already does this](https://github.com/tokio-rs/tokio/blob/master/.github/workflows/ci.yml#L184).

I was trying to do this for my crate that has an optional dependency on async-graphql, and hit a couple of compilation failures during this check. I can get around them by manually specifying transient dependencies in my build, but it's a good practice to try to contribute back to upstream dependencies where possible. 😄 

Given the large scope of async-graphql, going to have to manually step through each project and bump versions or add manual inclusion of transient dependencies until we're good to go.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "async-graphql"
-version = "2.7.3"
+version = "2.7.4"
 authors = ["sunli <scott_s829@163.com>", "Koxiaet"]
 edition = "2018"
 description = "A GraphQL server library implemented in Rust"
--- a/integrations/actix-web/Cargo.toml
+++ b/integrations/actix-web/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "async-graphql-actix-web"
-version = "2.7.3"
+version = "2.7.4"
 authors = ["sunli <scott_s829@163.com>", "Koxiaet"]
 edition = "2018"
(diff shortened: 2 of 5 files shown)


chipsenkbeil:
Also, you can use `cargo tree --all-features --workspace -p trust-dns-resolver:0.19.2 -i trust-dns-resolver` and similar to do a reverse lookup of what is bringing in specific versions of dependencies that are failing due to older versions. For example, the above could be updated to 0.19.6 following semver to possibly add the missing functionality.

Screen Shot 2021-01-15 at 12 55 06 PM

sunli829:
Thank you very much, I will merge it after I solve these problems. 🙂

chipsenkbeil:
@sunli829, polyval seems to be giving me some trouble finding a proper version of the dependency that works. I'll need to go back and see what version is pulled in during compilation without using the minimal-versions feature. Going to take a break for a little bit. 😄

PR #395, opened 2021-01-19
Book fix for `default_with`

leebenson:
Small fix for `default` -> `default_with` under "Default value" in the book, where a function is called.

Signed-off-by: Lee Benson <lee@leebenson.com>


sunli829:
Haha,thank you!

Issue #149, opened 2020-06-04
Question about #[SimpleObject] and #[InputObject]


tonis2:
I would like to to use `#[SimpleObject]` and `#[InputObject]` on a same struct, so i could use same data for query's parameters and output.

For example i would like to build a query

`fn filter_data(&self, data: DataSchema) -> Vec<DataSchema> {`

   _here_ i filter database entries based on data params and spew out the _results.._
`}`

Thanks

sunli829:
Currently, only two structures can be defined.

tonis2:
Okay thanks, using Json(DataSchema) as input type helped me out too.

AurelienFT:
Hello,

Keep two different structures is a choice or it will evolve in the future to allow us  to have one structure ?
@sunli829

sunli829:
I thought about it, I can support this feature, I am trying.

AurelienFT:
This will be awesome because Juniper do not support it at all.

zicklag:
@tonis2 could you explain a bit more about how you used `Json(DataSchema)` to help you problem? I'm in the same situation, but I don't understand exactly what type you specified for the input type.

zicklag:
I'm experimenting with a way to automatically generate the input and output types with an attribute macro here:

https://github.com/katharostech/async_graphql_extras

```rust
/// Information about the user
#[graphql_object(
    // Custom doc string for generated `UserDataInput` struct
    input_object_doc = "Input for user information data",
)]
pub struct UserData {
    username: String,
    display_name: String,

    // Indicate that this is also a `graphql_object`
    #[graphql_object(nested)]
    favorites: UserFavorites,
}

/// A user's favorite stuff
#[graphql_object]
pub struct UserFavorites {
    food: String,
}

#[Object]
impl Query {
    /// Ping endpoint that returns the same object as the input
    async fn ping(&self, user_input: UserDataInput) -> UserData {
        user_input.into()
    }
}
```

Rather simple and straight-forward for now. I'm trying to see if I can also automatically generate an input type that can be used to patch fields so that you can trivially create an "update" endpoint that allows you to pass only the fields that you want changed.

Acidic9:
This feature would be great! My project is quite dependant on it :(

PR #420, opened 2021-02-23
Add Note About Using Same Struct for In/Out

zicklag:
Added some documentation to the book that covers a workaround for #149.

### DIFF
--- a/docs/en/src/define_input_object.md
+++ b/docs/en/src/define_input_object.md
@@ -27,6 +27,37 @@ impl Mutation {
 }
 ```
 
+## Using the Same Struct for `InputObject`s and `SimpleObject`s
+
+It is not currently possible to derive `InputObject` and `SimpleObject` on the same struct ( see [#149] ). In some cases, though, it may be desireable to use the same Rust struct to represent both the input and output types. In this case you can use the [`Json`] struct as a wrapper for the input type, and then derive `SimpleObject` on the output type like normal. For instance:
+
+```rust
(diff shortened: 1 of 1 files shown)


sunli829:
Sorry, I don't think this is a good way, so I can't merge it.

It is not possible to implement both `SimpleObject` and `InputObject` for a `struct` due to the following two reasons:

1. Since each type can only implement `Type trait` once, the type information created by the `Type::create_type_info` function can only be `Object` or `InputObject`.

2. The field attributes of `SimpleObject` and `InputObject` are different.

sunli829:
There is a possible solution:

```rust
#[derive(SimpleObject)]
#[graphql(input="MyObjInput")]
struct MyObj {
     ...
}
```

This macro can generate a new `MyObjInput` structure, but the problem is that this structure cannot be recognized by the IDE, causing the editor to fail to complete it automatically.

zicklag:
> Sorry, I don't think this is a good way, so I can't merge it.

Yeah, I get it, that makes sense. :+1:

> There is a possible solution:

Yep, that's what I was thinking. I actually just did something like that yesterday: https://github.com/katharostech/async_graphql_extras.

It's was just some experimenting, but it worked fairly well. Obviously it would be better to have built-in. :smile: 

About auto-completion, I thought that the proc macro feature of rust analyzer was supposed to take care of that, but for some reason it isn't working with my example crate. :thinking: I might have to ask them about that.

PR #433, opened 2021-03-06
Update apollo_tracing.md

taiki-e:
`SchemaBuilder::extension` no longer takes a closure as an argument.


sunli829:
Thank you so much!😄

PR #434, opened 2021-03-07
Fix format of InputObject in SDL

taiki-e:
Currently, the format of Input object types in SDL returned by Registry::export_sdl is a bit odd because it doesn't insert tabs before the fields and their docs.

```rust
use async_graphql::*;

#[derive(Default)]
struct QueryRoot;
#[Object]
impl QueryRoot {
    async fn query(&self, input: Input) -> i32 {
        todo!()
    }
}
#[derive(InputObject)]
struct Input {
    /// doc
    field: i32,
}

type MySchema = Schema<QueryRoot, EmptyMutation, EmptySubscription>;

let schema = MySchema::default();
println!("{}", schema.sdl());
```

Before(master):

```graphql
type QueryRoot {
        query(input: Input!): Int!
}
input Input {
"""
doc
"""
field: Int!
}
schema {
        query: QueryRoot
}
```

After(this PR):
```graphql
type QueryRoot {
        query(input: Input!): Int!
}
input Input {
        """
        doc
        """
        field: Int!
}
schema {
        query: QueryRoot
}
```

### DIFF
--- a/src/registry/export_sdl.rs
+++ b/src/registry/export_sdl.rs
@@ -200,9 +200,9 @@ impl Registry {
                 writeln!(sdl, "{{").ok();
                 for field in input_fields.values() {
                     if let Some(description) = field.description {
-                        writeln!(sdl, "\"\"\"\n{}\n\"\"\"", description).ok();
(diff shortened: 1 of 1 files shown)


sunli829:
Thank you so much!😄

PR #438, opened 2021-03-09
Support generics in Subscription types

taiki-e:
Currently, defining a `Subscription` that contains generics causes errors.

```rust
struct MySubscription<T> {
    // ...
}

#[Subscription]
impl<T: OutputType + Type> MySubscription<T>
where
    T: Clone + Send + Sync + Unpin,
{
    async fn values(&self) -> Result<impl Stream<Item = T> + '_> {
        // ...
    }
}
```

```text
error[E0412]: cannot find type `T` in this scope
   --> tests/generic_types.rs:261:47
    |
261 |     impl<T: OutputType + Type> MySubscription<T>
    |                                               ^ not found in this scope
```


sunli829:
Thank you so much, works like a charm, haha. 😁

PR #457, opened 2021-03-30
Update rocket integration with new FromForm query parser

Punie:
Even though Rocket 0.5.0 has not been released yet, the big `Form` refactor landed on the master branch and with it [the way that query strings are handled](https://rocket.rs/master/guide/requests/#query-strings).

I upgraded the `async-graphql-rocket` integration crate with the latest revision of Rocket and implemented the query parser in terms of [`FromForm`](https://api.rocket.rs/master/rocket/form/trait.FromForm.html).

I originally attempted to implement `FromForm` manually for `async_graphql_rocket::Request` but it proved quite troublesome.
I then figured that a graphql query string can only have fields for `query`, `operationName` and `variables` and simply created a struct that could parse those by deriving `FromForm` and then converting that to a `async_graphql_rocket::Request`.

As it is my first contribution, feedbacks are very welcome.

**TODO:** update example to reflect the API change


sunli829:
Thank you so much, it would be better if you can help me update the examples, because I don't know how to use `Rocket`. 😁

PR #472, opened 2021-04-03
Document SDL schema export in the book

tobisinghania:
This pull request adds an example to the book of how the code-first schema can be exported as SDL.

I was looking for a way to export the schema as SDL and it took me some time to find out how it's done, because this feature is mentioned nowhere in the docs.


tobisinghania:
Seems CI is failing because a diff with `cargo fmt` results in a file which I have not modified (`subscription.rs`).

If I should run `cargo fmt` on my end and recreate the pull request, then let me know.

sunli829:
LGTM,thanks!😄

PR #475, opened 2021-04-05
Update cursor_connections.md

dsonck92:
Add the necessary `async move` to the block.

This actually took me way too long to realize. Granted, I'm not yet well versed in Rust so I had a hunch this was required, but did not yet know where it was supposed to go.


sunli829:
Thank you, the documentation is wrong. I have `cherry-pick` your commit, so I'll close this PR.

PR #483, opened 2021-04-17
add missing fonts to http::playground_source's html

utkarshkukreti:
Hi. Thanks for creating this library. I was playing around with it and found the typography a bit off in the playground. Turned out the playground was using a font that was not being included by the code in this crate.


sunli829:
Thanks.

PR #485, opened 2021-04-19
Add link to license files

LiteLT:
A small request, but I think it's a bit awkward to have the path to the license files in `README.md` but not have them clickable. This commit adds links to the licenses.


sunli829:
Thanks.

PR #496, opened 2021-05-04
Improve documentation of ComplexObject

bram209:
Some small improvements in the documentation of `ComplexObject`.
I think we could further improve it by having a dedicated `ComplexObject` page as I think the use case of a complex object is quite common.

The graph nature of graphql makes that you would often need `ComplexObject` to resolve the relationships of an entity.
Next to demonstrating the use case of calculated fields, it would maybe be a good idea to add another example that fetches a relationship?

On a related note, think the example of https://async-graphql.github.io/async-graphql/en/dataloader.html#dataloader would be better fitted to showcase a data loader on the `User` graphql object as a whole instead of a single field from the database.

Thanks for the lib, has been a joy to work with!


sunli829:
Thank you so much! 😁

PR #506, opened 2021-05-12
improve guard and

zxk7516:


sunli829:
Thanks, it looks good.

SirWindfield:
Out of curiosity: wouldn't it be more efficient to use `or_else` to lazily evaluate the second statement?

PR #508, opened 2021-05-14
Allow the ability to set GraphQL Playground settings

oscartbeaumont:
In the GraphQL Playground UI, you can configure settings. This PR allows a default set of settings to be defined server-side similar to the default headers option that already exists in this package.

An example use case of this feature is setting the `request.credentials` setting to `include` or `same-site` instead of its default value `omit`. This allows cookies to be manipulated in the GraphQL request.

### DIFF
--- a/src/http/playground_source.rs
+++ b/src/http/playground_source.rs
@@ -596,15 +596,15 @@ impl<'a> GraphQLPlaygroundConfig<'a> {
         self
     }
 
-     /// Set Playground setting for per query.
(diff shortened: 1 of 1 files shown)


sunli829:
Thank you so much! 🙂

oscartbeaumont:
I have linted the code as that was causing the workflow failures. Hopefully, we can get this merged!

sunli829:
Thank you, I forgot to merge it. 🙂

PR #516, opened 2021-05-19
feat: support bson v2.0.0-beta

flisky:
This could be merged after the final release :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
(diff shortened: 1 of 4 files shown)


sunli829:
When `bsonv2` is finally released, I think we can remove support for `bsonv1`, what do you think?

flisky:
I'm okay with that, but it should bump the version due to a breaking change?

AurelienFT:
To add bson recommend 2.0.0-beta.1 in their README.md : https://github.com/mongodb/bson-rust

flisky:
Aha, thank you for reminding me that bson -> mongo -> tokio dependency chain, so it's perhaps a expected breaking change. I'll update this PR shortly.

flisky:
Could you please take another look, @sunli829? Thanks!

sunli829:
Thank you so much! 🙂

AurelienFT:
@sunli829 Can you make a feature realease so that we can use it in our project please ? :) And thanks again for your reactivity and your kindness with your community

AurelienFT:
Released in 2.9.3.

PR #529, opened 2021-05-28
Add WSSubscription drop callback

fayalalebrun:
For the application I am working on, I need to know whether a client has lost its WS connection to the server in order to handle it accordingly.

For this reason, I implemented an extra field in the `WSSubscription` object, which is a callback called whenever the object is dropped. The reason I did not call this callback when the close message is sent instead of drop, is because if a client crashes, then the close message will not be sent. However, the object will always be dropped, no matter how the connection is closed.

### DIFF
--- a/integrations/actix-web/src/subscription.rs
+++ b/integrations/actix-web/src/subscription.rs
@@ -27,6 +27,7 @@ pub struct WSSubscription<Query, Mutation, Subscription, F> {
     messages: Option<async_channel::Sender<Vec<u8>>>,
     initializer: Option<F>,
     continuation: Vec<u8>,
+    drop_handler: Option<Box<dyn FnOnce()>>,
 }
 
 impl<Query, Mutation, Subscription>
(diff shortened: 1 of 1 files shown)


sunli829:
Thank you for the PR, but I don't think this implementation is elegant enough, although it does look very convenient to use. Please give me some time, and I'll think about how to better achieve this goal.

fayalalebrun:
I think so as well. For the constructor, I think the best would be to use a builder pattern, though I didn't do this as it would break the code of anyone using this.

As for using the Drop trait to achieve this, perhaps there is a better way, though I'm not familiar enough with the actix internals to tell.

sunli829:
I think the following code can be done, if you still don’t understand, I’m happy to provide you with a simple example.

```rust

/// stream wrapper
struct MyStream(T);

/// handle drop
impl Drop for MyStream {
    ....
}

impl Stream for MyStream {
    type Item = actix_web::Result<Bytes, PayloadError>;

    fn poll_next(...) {
         self.0.poll_next(cx)
     }
}

// start the websocket actor
WSSubscription::start(schema, &req, MyStream(payload));
```

PR #535, opened 2021-06-02
Aligned NaiveDateTime formatting with DateTime

### LINKED ISSUE #526: Inconsistency between NaiveDateTime and DateTime formatting
Currently the date formatting used by `NaiveDateTime` (and `NaiveTime`) is inconsistent with `DateTime`.

A `NaiveDateTime` stamp looks like: `2021-05-27 01:30:32`, while `DateTime` looks like: `2021-05-27T01:30:32.747575+00:00`.

Notice the `T` over the space and milliseconds. As my API uses both, for consistency I feel it'd make more sense to change `NaiveDateTime`'s [implementation](https://github.com/async-graphql/async-graphql/blob/master/src/types/external/naive_time.rs#L43) to more closely follow the RFC3399 [implementation](https://github.com/async-graphql/async-graphql/blob/master/src/types/external/datetime.rs#L52) `DateTime` uses.

I understand that technically `NaiveDateTime` still wouldn't be RFC3399, but given the only difference to an end user between the two is the presence of a timezone aligning them makes more sense in my eyes.

I'd be happy to make a PR if desired, although I think it's important to note that this would be a breaking change as parsing `NaiveDateTime` currently expects a space.

oeed:
Fixes #526

Chrono uses `%Y-%m-%dT%H:%M:%S%.f%:z` for it's RCC3339 formatting implementation. See https://github.com/chronotope/chrono/blob/3467172c31188006147585f6ed3727629d642fed/src/format/mod.rs#L680

### DIFF
--- a/src/types/external/naive_time.rs
+++ b/src/types/external/naive_time.rs
@@ -20,13 +20,13 @@ impl ScalarType for NaiveDate {
 impl ScalarType for NaiveTime {
     fn parse(value: Value) -> InputValueResult<Self> {
         match value {
-            Value::String(s) => Ok(NaiveTime::parse_from_str(&s, "%H:%M:%S")?),
+            Value::String(s) => Ok(NaiveTime::parse_from_str(&s, "%H:%M:%S%.f")?),
             _ => Err(InputValueError::expected_type(value)),
         }
     }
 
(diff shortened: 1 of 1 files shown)


sunli829:
Thanks! 😁

PR #559, opened 2021-07-01
Add Decimal scalar (from 'rust_decimal' crate)

bram209:
When your application has to deal with financials (i.e. billing/accounting/invoicing) you often need to deal with decimals.
Needed it for my use case so thought a `decimal` feature will be handy.

I prefer it being supported by `async-graphql` natively.  sqlx supports a similar feature, meaning I can conveniently query decimals from my DB and serve it through graphql api directly without needing a custom newtype and scalar.


sunli829:
Thank you so much! 😁

PR #561, opened 2021-07-03
Remove outdated documentation regarding subscriptions in Tide

okrplay:
The Tide integration documentation still said that subscriptions are not supported due to missing websocket support in Tide. As subscriptions are supported by now, this note seems outdated and misleading.


sunli829:
Thanks!

PR #562, opened 2021-07-03
#550: macros fixes for generic Context

### LINKED ISSUE #550: [Question] Generics and context
Hello 👋 

First of all thank you for this amazing crate. I'm having issues achieving a particular use case and I'm curious if its at all possible or I'm looking at the problem from the wrong perspective.

In my application I have service structs that look like this:

```rust
#[derive(Clone)]
pub struct UserService<T: UserRepository> {
    repo: T,
}
```

From there I can initialize `UserService` with a `PostgresRepo` or a `InMemoryRepo`, which I use in tests.

**Now I hit a wall when trying to keep this generic while using services through the GraphQL context.**

```rust
    Schema::build(Query, EmptyMutation, EmptySubscription)
        .data(user_service)
```

This works fine, but it's when I want to use these services from resolvers that things get tricky:

```rust
        let user_service = ctx
            .data::<UserService<CanIKeepThisGeneric?>>()
            .expect("Missing user service from context");
```

It seems like at this point I must use a concrete type ☝️ I've tried multiple approaches but could not find a way to keep resolvers generic which would allow me to test my schema with mocks. Right now I've opted for `UserService<PgUserRepository>` and plan to test resolvers in full integration tests only. Would there be a way to achieve this?

nicoulaj:
This fixes 2 small bugs in `ComplexObject` to allow generic typing of `Context` data. See unit test for example.

Fixes #550

### DIFF
--- a/derive/src/complex_object.rs
+++ b/derive/src/complex_object.rs
@@ -28,6 +28,7 @@ pub fn generate(
             let method_args: args::ObjectField =
                 parse_graphql_attrs(&method.attrs)?.unwrap_or_default();
             if method_args.skip {
+                remove_graphql_attrs(&mut method.attrs);
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


sunli829:
Thank you, I also added support for concrete types of `ComplexObject`.

PR #564, opened 2021-07-08
Feature: Add Error with Extension in Validator

cptrodgers:
Hi guys, this is my first contribution, please correct me.
My PR to help error data of input validator can contain more information by `ErrorExtentionValues`.

There are some use cases that I can imagine:
- Use code number or code string for show exactly error (Bad Request, Internal Server Error) on Client Side.
- Support localization (i18n) by adding key and its data on extensions.
- ...

Example implementation:
```rust
    use async_graphql::*;
    pub struct IntGreaterThanZero;

    impl InputValueValidator for IntGreaterThanZero {
        fn is_valid_with_extensions(&self, value: &Value) -> Result<(), Error> {
            if let Value::Number(n) = value {
                if let Some(n) = n.as_i64() {
                    if n <= 0 {
                        let e = Error::new("Value must be greater than 0")
                            .extend_with(|_, e| e.set("code", 400));
                        return Err(e);
                    }
                }
            }
            Ok(())
        }
    }
```

Expected Output:

```json
{
  "data": null,
  "errors": [
    {
      "message": "... Value must be greater than 0",
      "locations": [],
      "extensions": { "code": 400 }
    }
  ]
}
```

Compatible note:
- Compatible with all logic code of fn is_valid.
- ~Break MapErr logic on `src/validators/mod.rs`. Closure from `fn(String) -> String` to `fn(Error) -> Error`~. Compatible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.9.8] 2021-07-09
+
+- Add Extensions in Error of `InputValueValidator`. [#564](https://github.com/async-graphql/async-graphql/pull/564)
+
 ## [2.9.7] 2021-07-04
 
 - Add support for generic `ComplexObject`. [#562](https://github.com/async-graphql/async-graphql/pull/562)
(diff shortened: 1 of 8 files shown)


cptrodgers:
cc @sunli829 . I'm updating the logic for `map`, `or`, and `map_err` but you can advise me.

sunli829:
Thank you very much, it looks great!
If you are free, please help update `CHANGELOG`. 🙂

cptrodgers:
> Thank you very much, it looks great!
> If you are free, please help update `CHANGELOG`. 🙂

Added. Thanks @sunli829 .

PR #574, opened 2021-07-15
Changed Lookahead to support multiple fields

### LINKED ISSUE #573: Lookahead doesn't allow multiple instances of a subfield
I've encountered a limitation to `Lookahead` that I think severely limits their usability, particularly when used in public facing APIs. For internal only APIs you can (sometimes) get around this if aware of the limitations, but I feel it'd be inadvisable to use them in their current state as it could cause a lot of unexpected behaviour.

Take a query with the following field selections:

```gql
myQuery {
  a {
    b
  }

  a {
    c
  }
}
```

While it might seem strange to do this, using fragments can commonly result in this sort of query and can be unavoidable without inlining the fragment content. For example, in React Apollo client it's common for components to have their own fragments. If component `B` needs field `b` and component `C` needs field `c`, for example.

```gql
myQuery {
  ...componentB
  ...componentC
}

fragment componentB on MyQuery {
  a {
    b
  }
}

fragment componentC on MyQuery {
  a {
    c
  }
}
```

In my resolver if I do `&ctx.look_ahead().field("a").field("c")`, it'll indicate that it doesn't exist because `.field("a")` simply takes the first `a { }`. Instead of a single `field: Option<&'a Field>`, I think `Lookahead` really needs to use `field: Vec<&'a Field>`.

I'd be quite happy to provide a PR to provide this. There might be some performance penalties I'm unaware of though, so let me know if so.

oeed:
Fixes #573

### DIFF
--- a/src/look_ahead.rs
+++ b/src/look_ahead.rs
@@ -6,7 +6,7 @@ use crate::{Name, Positioned, SelectionField};
 /// A selection performed by a query.
 pub struct Lookahead<'a> {
     fragments: &'a HashMap<Name, Positioned<FragmentDefinition>>,
-    field: Option<&'a Field>,
+    fields: Vec<&'a Field>,
 }
 
 impl<'a> Lookahead<'a> {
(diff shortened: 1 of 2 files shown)


src/look_ahead.rs
```diff
@@ -16,63 +16,68 @@ impl<'a> Lookahead<'a> {
...
     fragments: &'a HashMap<Name, Positioned<FragmentDefinition>>,
     selection_set: &'a SelectionSet,
     name: &str,
-) -> Option<&'a Field> {
-    selection_set
-        .items
-        .iter()
-        .find_map(|item| match &item.node {
+) {
+    for item in &selection_set.items {
```
oeed, line 62: I've done this imperatively. Tried doing this with iterators but I'm pretty sure you need to use a `Box<dyn Iterator>` because it's recursive. Happy to change if there's a better way though.

sunli829:
You are right, I really didn't think about this. 👏🏻

sunli829:
If you are free, please help update CHANGELOG. 🙂

oeed:
No worries, have added

sunli829:
Thank you so much! 😁

Issue #557, opened 2021-06-30
Lookahead from SelectionField


oeed:
For 99% of my use cases I just need to traverse the tree of fields and know if a certain name is present. I currently do this using `Lookahead::exists` and `Lookahead::field`.

However, for certain fields I need to iterate the _aliases_ (a field with one alias needs to be treated differently to another with the same name as they can have different arguments). I can do this using `SelectionField::selection_set` and checking if the name is the field I need to handle.

After I've got the alias field I then need to use the `Lookahead` again for the subfields that fall in to the 99% use case. I can't convert between these two myself as the internal values are private, but it looks like it should be straight forward to have an `impl From<SelectionField> for Lookahead`.

Alternatively, a way to iterate all aliases on `Lookahead` would be ideal.

sunli829:
`SelectionField` and `LookAhead` are used to do the same thing, but provide more information, such as field name, alias, etc.

oeed:
I imagine there is a good reason between why there are two separate versions, let me illustrate my situation just so we're on the same page. Say I have the following schema:
```gql
type Parent {
  child(id: Int!): Child!
}

type Child {
  name
}

type Query {
  parent(id: Int!): Parent!
}
```

Then I query:
```gql
{
  parent(id: 1) {
    alice: child(id: 2) {
      name
    }

    bob: child(id: 2) {
      name
    }
  }
}
```

In the query resolver for `Parent` say I need to fetch all data including all children in one query, rather than fetching children separately (this is my situation and is needed for various reasons). In order to do so I need to know all `child` fields and their IDs.

I can do this using `SelectionField` and iterating `selection_set` for where `name == "child"` and get the ID and alias.

Then, each `Child` needs to know what fields it needs to fetch to build its part of the query. With `Lookahead` this is easy:
```rs
lookahead.field("name").exists()
```
However, because I found `alice` and `bob` using `SelectionField` I cannot get a `Lookahead`. `Lookahead::field` simply gets the first field with that name, I cannot get `Lookahead` for `alice` and `bob` separately. Therefore I _must_ use a `SelectionField`.

`SelectionField` doesn't have any `exists` or `has_field` type of function - instead I have to manually iterate `selection_set` and find a match, which is cumbersome and doesn't allow me to reuse the field -> query code (in other `Child` queries I might need `Lookahead`).

Basically, at the moment you have to pick to use the same struct everywhere, not a combination of the two. Does that illustrate the problem I'm facing?

---

[...]

PR #575, opened 2021-07-15
Implement `TryFrom<&[SelectionField<'a>]>` for `Lookahead<'a>`

oeed:
Sorry this really should have been part of #574, this adds support to do what #557 mentioned but for multiple `SelectionField<'a>`. It does assume `fragments` are the same for all, I presume `fragments` is the same for everything in the same query and therefore it's okay to assume this?


sunli829:
Thanks! 😁

PR #578, opened 2021-07-18
Allow field visible to support paths

flisky:
I think it's good to have, more consistent with others, and less surprise to users.


sunli829:
Looks great, thank you!

PR #584, opened 2021-07-23
Change with_setting to accept any JSON value

### LINKED ISSUE #583: Playground settings can only be &str
## Expected Behavior

Playground settings should be able to accept any JSON type (although maybe not Object or Array?).

## Actual Behavior

[`with_setting`](https://docs.rs/async-graphql/2.9.9/async_graphql/http/struct.GraphQLPlaygroundConfig.html#method.with_setting) can only accept a `&str` for the value, so any values passed are serialised into a JSON string. These then stay as a string while being inserted into the playground settings.

```rust 
playground_source(
      GraphQLPlaygroundConfig::new("/admin/api/graphql")
          .with_setting("schema.polling.enable", "false"),
)
```

image

## Specifications

- Version: 2.9.9
- Platform: Linux

I think this could be solved by changing `GraphQLPlaygroundConfig`'s settings Hashmap into `Hashmap<&'str, Value>` and then changing `with_setting` to accept a Value. 

I'd be happy to put a PR together for this, let me know if you think this is viable.

tim-maddison-rnf:
Closes #583

### DIFF
--- a/src/http/playground_source.rs
+++ b/src/http/playground_source.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use crate::Value;
 use serde::Serialize;
 
 /// Generate the page for GraphQL Playground
@@ -564,7 +565,7 @@ pub struct GraphQLPlaygroundConfig<'a> {
     endpoint: &'a str,
     subscription_endpoint: Option<&'a str>,
     headers: Option<HashMap<&'a str, &'a str>>,
-    settings: Option<HashMap<&'a str, &'a str>>,
+    settings: Option<HashMap<&'a str, Value>>,
 }
 
 impl<'a> GraphQLPlaygroundConfig<'a> {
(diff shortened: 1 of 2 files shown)


tim-maddison-rnf:
I haven't added anything to the changelog, let me know if you want me to do it.

sunli829:
> I haven't added anything to the changelog, let me know if you want me to do it.

Yes, please do it. 😁

src/http/playground_source.rs
```diff
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use crate::Value;
```
sunli829, line 3: Please modify it as follows: 😁

```rust
use std::collections::HashMap;

use serde::Serialize;

use crate::Value;
```

sunli829:
Thank you!👏

tim-maddison-rnf:
And thank you for working on async-graphql, I've only been using it for about a week, but it's been an absolute pleasure to work with.

PR #594, opened 2021-08-03
Remove unnecessary `unwrap` in multipart handler

nicolaiunrein:
I experienced panics when the client drops the connection while a file is being uploaded. Turns out the current implementation of the multipart handler uses an `unwrap` for no good reason. This PR replaces the call to `unwrap` and instead returns a `ParseRequestError::InvalidMultipart(multer::Error)`.


sunli829:
This is really a serious problem. 😂

PR #600, opened 2021-08-13
Implement `From<MaybeUndefined<T>> for Option<Option<T>>`

### LINKED ISSUE #599: Implement From<MaybeUndefined<T>> for Option<Option<T>>
Hey @sunli829 ! Thanks for all the work you put into this crate. It's truly awesome.

## Description of the feature

I think it would be useful to easily convert from `MaybeUndefined<T>` to `Option<Option<T>>`,
where `Undefined => None`, `Null => Some(None)` and `Value(a) => Some(Some(a))`.

I'm using the `Option<Option<T>>` pattern to tell in my SQL queries if I should set the value to null or to not update it at all (custom code in [SQLx](https://github.com/launchbadge/sqlx)), and Diesel [uses this pattern by default](https://diesel.rs/guides/all-about-updates.html#aschangeset) too.

## Code example

This is the main type of use case I'm imagining:

```rust
struct UserChangeset {
  pub name: Option<Option<String>>,
}

#[Object]
impl Mutation {
  async fn change_name(&self, ctx: &Context<'_>, name: MaybeUndefined<String>) -> Result<()> {
    // ...
  
    user.update(UserChangeset { name: name.into() }).await
  }
}
```

What do you think? (I can implement the feature if you accept it)

dbstratta:
Closes #599

### DIFF
--- a/src/types/maybe_undefined.rs
+++ b/src/types/maybe_undefined.rs
@@ -156,6 +156,16 @@ where
     }
 }
 
+impl<T> From<MaybeUndefined<T>> for Option<Option<T>> {
+    fn from(maybe_undefined: MaybeUndefined<T>) -> Self {
+        match maybe_undefined {
+            MaybeUndefined::Undefined => None,
+            MaybeUndefined::Null => Some(None),
+            MaybeUndefined::Value(value) => Some(Some(value)),
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


sunli829:
Thank you, this seems very useful!  😄

Issue #603, opened 2021-08-17
2.9.8 started serializing my uuid's as arrays instead of strings


stijnfrishert:
I have a custom newtype that contains a [`Uuid`](https://docs.rs/uuid/0.8.2/uuid/), that basically looks like this:

```
#[derive(Serialize, Deserialize)]
#[repr(transparent)]
#[serde(transparent)]
AssetId(pub uuid::Uuid)
```

This is then wrapped into an `async_graphql::scalar!(AssetId)`.

- In versions below `2.9.8` this used to serialize (to JSON) as a hyphenated string, such as `"3292da00-ff2c-45aa-9d8e-741f6a3b0b0c"`.
- Starting from `2.9.8`, this is serialized as a byte array, such as `[113, 225, ...]`.
- Looks like the `Serializer` used in `2.9.8` isn't human readable, which leads `Uuid` to use the byte array serialization.

What changed? And more importantly, why did that change? `CHANGELOG.md` doesn't mention anything about Uuid's or serialization, and this seems like a pretty big breaking change (warranting a major version upgrade).

Miaxos:
You sure it's the `2.9.8` which is not working for your use case?

I tried to create a reproductible use case, and from my findings, it's this commit from the https://github.com/async-graphql/async-graphql/commit/ec8f97ba9e1c37433742dc5b8f9df1a02dfb9ec9  (2.9.9 release) which is failing with the implementation of `Deserialize` and `Serialize` instead of the trait auto-generation https://github.com/async-graphql/async-graphql/commit/ec8f97ba9e1c37433742dc5b8f9df1a02dfb9ec9#diff-592c2b8e5e98ef9b09e52148d90a976f0153e3f0659a0489725f0ce26669da3aL124-L125

The test case https://github.com/async-graphql/async-graphql/pull/604

stijnfrishert:
Double checked for you @Miaxos, but looks like you're right. I switch between versions `=2.9.7` and `=2.9.8`, assuming I'd get exactly [the specified version](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility), but looking at the `Cargo.lock` it jumped up to `2.9.10`.

Thanks for taking a look so quickly. Sounds like CBOR won't be a problem as it shouldn't be `human_readable`, but I'll leave that to the pro's ;)

Erik1000:
> Are you sure it's the `2.9.8` which is not working for your use case?
> 
> 
> 
[...]

PR #604, opened 2021-08-18
FIX: Serializing issue

Miaxos:
Test case for serializing issue https://github.com/async-graphql/async-graphql/issues/603


sunli829:
Thank you so much!😄

PR #605, opened 2021-08-19
Ignored items flagged @skip in SelectionField and Lookahead

oeed:
Currently if you have a query with a field/fragment marked as `@skip(if: true)` or `@include(if: false)` it'll still be included by `SelectionField` and `Lookahead`. As they're used to determine what was requested they should be excluded in my opinion.

This PR simply ignored any skipped fields/fragments. Before merging, it's worth considering what to do on `is_skip` returns `Err`. For now I'm ignoring the error, as I'm assume it'll always be caught elsewhere anyway and it would introduce backwards compatibility issues having to handle `field("name")?` everywhere; but I think that should be discussed.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -603,7 +603,7 @@ impl<'a> ContextBase<'a, &'a Positioned<Field>> {
     /// }
     /// ```
     pub fn look_ahead(&self) -> Lookahead {
-        Lookahead::new(&self.query_env.fragments, &self.item.node)
+        Lookahead::new(&self.query_env.fragments, &self.item.node, self)
     }
 
     /// Get the current field.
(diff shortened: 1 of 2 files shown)


Miaxos:
Yep it should throw later with the `add_field`, but if the execution is changed later and it do not throw anymore it might create unwanted side effects to just have the `unwrap(false)` later no?

If I may say so, here is my opinion: I like the idea of throwing early but this requires a major version as you said

sunli829:
Let me think about it, maybe there is a way to prevent `Lookahead` from getting the skipped fields.

sunli829:
I didn't think of a more elegant way. Ignoring the error message of `is_skip` will not cause trouble, because other resolvers will find this error.

Thank you so much for doing this! @oeed 😁

sunli829:
These are no longer needed in the `v3` version, because these fields will be removed before execution.

Issue #569, opened 2021-07-12
Binary Transmission


cloudybyte:
We want to transmit a `Vec<u8>`. Because JSON does not support binary data we chose CBOR for transmittion, but it appears that binary transmission is [not supported](
https://docs.rs/async-graphql-value/2.9.8/src/async_graphql_value/lib.rs.html#126-142) in async-graphql? Or are we looking at the wrong place?

sunli829:
Binary transmission requires you to implement it yourself, which is easy. You just need to `Response` serialized as binary format you want.

cloudybyte:
I think you misunderstood what I meant. We want to transmit data like images and pgp keys...

cloudybyte:
We basically want to encode the traffic in CBOR and use custom scalars to send binary data in the encoded response

sunli829:
Ah, I see, you want `Value` to support binary types, so that CBOR can serialize binary values.

sunli829:
I implemented this on the master branch, I hope you can help test it.

cloudybyte:
We're currently trying to get cbor serializing to work in order to test the binary transmission, but aren't able to test our implementation because graphiql and graphql playground both don't support deserializing it...

cloudybyte:
it seems like we can't just serialize the request in cbor, because the deserializing is done in async-graphql. I tried to implement a function, but it seems like the errors depending on it are hardcoded to `serde_json::Error`

Miaxos:
Maybe if we could change the args to a dynamic impl of Error ?
https://github.com/async-graphql/async-graphql/compare/master...Miaxos:master

cloudybyte:
We tried that, but that was causing many more problems... If it compiles and works I guess it's a good fix

Miaxos:
>but that was causing many more problems

Can you elaborate on the problems?

cloudybyte:
Basically just causing many 'mismatched type' errors

Miaxos:
ok, it seems good with my change

Erik1000:
Got it working for incoming cbor on [our branch](https://github.com/minkan-chat/async-graphql).

However, the whole way it is implemented doesn't look good to me. For the library I am thinking to decide which method to use based on the ``Accept-Encoding`` header. E.g. if it is set to ``application/json`` return json or if ``application/cbor`` return cbor. Also most if not all of the tests fail because of the strict use of serde_json. @sunli829 what do you think?

sunli829:
[...]

PR #619, opened 2021-09-02
Add support for cbor

Erik1000:
A while back we discussed the ability to use cbor as encoding. Back then, we got that to work and used it since on our fork. I thought that we could make a PR too.

# Note

We use it to build a graphql server and haven't encountered any errors *yet*. Multipart (for the file upload extension) probably won't work just yet, because it still uses serde_json.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ pin-project-lite = "0.2.6"
 regex = "1.4.5"
 serde = { version = "1.0.125", features = ["derive"] }
 serde_json = "1.0.64"
+serde_cbor = "0.11.1"
 thiserror = "1.0.24"
 static_assertions = "1.1.0"
 http = "0.2.3"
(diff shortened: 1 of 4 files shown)


Erik1000:
Related issue: https://github.com/async-graphql/async-graphql/issues/569

sunli829:
Thank you very much, it looks great! 🙂

Erik1000:
forgot to format that i guess

sunli829:
> forgot to format that i guess

I will solve it, and I will add a feature called `cbor`, what do you think?

Erik1000:
> I will solve it, and I will add a feature called cbor, what do you think?
seems like a good idea :+1:

Erik1000:
Should I merge your master into our fork so clippy passes(see latest commit comment)?

sunli829:
Yes, you can deref `Bytes` to `&[u8]`.

https://docs.rs/futures/0.3.17/futures/io/trait.AsyncRead.html#impl-AsyncRead-for-%26%27_%20%5Bu8%5D

sunli829:
I have merged into the `master`.

sunli829:
> Also I'm currently looking into the multipart code and there's a problem with multer:
I want to reduce code duplication so here I want to reuse the receive_batch_body method.

You can create a new PR to continue to improve this issue. 🙂

Erik1000:
I'll do :+1:

PR #629, opened 2021-09-10
Added Axum error handling

emeryc:
in this case the todo! was eating errors that I needed to see in order to debug my own issues.

Hopefully this is a reasonable implementation of the error handling for Axum


sunli829:
Hahaha, I forgot to write the code here, thanks you so much! 😁

PR #630, opened 2021-09-11
Add document for Input Validator with Extensions

cptrodgers:
Just an update example for `InputValueValidator` with Extensions in async doc.


sunli829:
Thanks!😃

Issue #637, opened 2021-09-19
Relay Persisted Queries


maraisr:
## Description of the feature

With the new Relay Rust compiler — it'll post queries during compile time. Would be great to offer an endpoint here to receive those FormData and persist to a backplane of your choosing.

## Code example (if possible)

https://github.com/facebook/relay/blob/b25725dd51e4cca8693dc018b7eaf25e81846441/compiler/crates/persist-query/src/lib.rs#L39

Miaxos:
Hello,

It seems a lot like the Apollo Persisted Queries: https://github.com/async-graphql/async-graphql/blob/master/src/extensions/apollo_persisted_queries.rs

I think we could create a similar extension to manage this relay feature.
I will have some time this week to look into it.

sunli829:
> Hello,
> 
> It seems a lot like the Apollo Persisted Queries: https://github.com/async-graphql/async-graphql/blob/master/src/extensions/apollo_persisted_queries.rs
> 
> I think we could create a similar extension to manage this relay feature.
> I will have some time this week to look into it.

Thank you! 🙂

maraisr:
Guess just pointing out, these queries are only posted once during their generation. So actually less like Apollo, in that it's not a cache. These queries need to persist somewhere cross runs.

May I suggest opening up an api to connect up Redis or alike?

Miaxos:
>Guess just pointing out, these queries are only posted once during their generation. So actually less like Apollo, in that it's not a cache. These queries need to persist somewhere cross runs.

From my understanding, two modes are possible, either posted within their generation, or with the run time push. https://relay.dev/docs/guides/persisted-queries/

Not sure how the runtime mode in relay is working, so I'll just implement the first one right now.

>May I suggest opening up an api to connect up Redis or alike?

Yes ofc! It should be possible to do it inside an extension easily and not inside the `async-graphql` engine.

maraisr:
> From my understanding, two modes are possible, either posted within their generation, or with the run time push. 

Ahhh that's missing it. With relay there is the existing javascript based compiler, that is documented in those docs. I was specifically speaking to the new rust based compiler, which isnt documented. 

But in anycase, when you ask the compiler to persist, it **WILL NOT** produce any query text — so there will be no query text to support the "apollo pattern". 

Relay is strictly generating the queries during build time — and persisting those then. At runtime, it ONLY has access to the id. No operation text. Both the old rust compiler, and the new one had this behavior.

Notice;

[...]

PR #641, opened 2021-09-20
feat: add relay persisted_query extension

Miaxos:
A draft to add relay persisted Query extension.

The purpose is to check if this could solve https://github.com/async-graphql/async-graphql/issues/637
If it's in the right way, test are to be added and the allowlist should also be implemented.

It's only slightly different than Apollo's version, so maybe merge them later?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ readme = "README.md"
 default = []
 apollo_tracing = ["chrono"]
 apollo_persisted_queries = ["lru", "sha2"]
+relay_persisted_queries = []
 unblock = ["blocking"]
 string_number = ["num-traits"]
 dataloader = ["futures-timer", "futures-channel", "lru"]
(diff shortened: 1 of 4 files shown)


maraisr:
Just double checking here — this isnt in query execution time. The query is uploaded via FormData well and truely before its execution time. 

At runtime, the query is in fact sent via the `id` property. Perhaps im missing the mark, but Apollo and Relay persistence layers are very different concepts.

Miaxos:
Yep you are right, I didn't understand you only wanted the endpoint to store the queries.

But for `async-graphql` to work when a relay client ask for a persisted query, you'll need an extension like this.

>At runtime, the query is in fact sent via the id property

The query is sent with the id property when compiling but with a `doc_id` from a relay client (from the documentation)
image

maraisr:
A consumer 100% controls the network layer — we can send whatever we'd like. `doc_id` is a facebook way, and 100% just for illustration. If async-graphql want to standardize on `id`, that is more than fine. Heck you could even make that field configurable.

maraisr:
hi @Miaxos  was wondering if you're still looking after this? If not, id be more than happy to fork your PR and help ship this across the finishing line 🚀

PR #643, opened 2021-09-26
Get selection fields from Lookahead

oeed:
In much the same was as it's useful to get `Lookahead` from `SelectionField` #557, the inverse is also quite useful.

This adds a function that generates all `SelectionField`s fo each field in the `Lookahead`. I decided to do this over implementing `From`, but I'm open to changing that.


oeed:
Yes good point! Done

sunli829:
Thank you! 😁

PR #646, opened 2021-10-03
docs(merging_objects): add missing structs in example

ecklf:
Hey there @sunli829,

Given the structs are included in the merging subscriptions example, I thought it might be nice to add them to the merging objects section for consistency.

Thanks for the library, works great.


sunli829:
Thank you so much!😀

PR #651, opened 2021-10-06
docs(default_value): fix macro syntax error

ecklf:
Fixes wrong macro syntax which caused the `InputObject` macro not to be displayed in the documentation.

<https://async-graphql.github.io/async-graphql/en/default_value.html#input-object-field>


sunli829:
Thanks!😄

sunli829:
> 很高兴能帮上忙 👍

哥们中文不错！😀

PR #655, opened 2021-10-11
feat: add visible macro argument for union type

Miaxos:
Hello 👋 !

A little PR to add a visible field parameter for Union types.

Was wondering how to make it, there should be a more optimized way to do this, but it's working well.


sunli829:
`2.10.3` has been released.

sunli829:
When I have completed the feature mentioned in #698, I will delete this because it is no longer needed.

PR #665, opened 2021-10-21
Fix the `Subscription` procmacro to work on Rust 2021 by adding the missing required `dyn` keywords in the generated output

### LINKED ISSUE #663: Rust 2021 Edition - Subscription procmacro fails and builds break due to lacking `dyn` keyword in generated output
## Expected Behavior

From the examples in the docs, this should work, and indeed in Rust 2018 edition it does work:

```rust
pub struct Subscription;

#[Subscription]
impl Subscription {
	async fn interval(&self, #[graphql(default = 1)] n: i32) -> impl Stream<Item = i32> {
		...snip...
	}
}
```

## Actual Behavior

Upon migrating to the Rust 2021 edition it fails with this error with the `impl Stream<Item = i32>` being highlighted in the error:

```rs
error[E0782]: trait objects must include the `dyn` keyword
   --> src/server/gql.rs:182:67
    |
182 |     async fn interval(&self, #[graphql(default = 1)] n: i32) -> impl Stream<Item = i32> {
    |                                                                      ^^^^^^^^^^^^^^^^^^
    |
help: add `dyn` keyword before this trait
    |
182 |     async fn interval(&self, #[graphql(default = 1)] n: i32) -> impl dyn Stream<Item = i32> {
    |                                                                      +++
```

Which is of course very not right, it looks like the `impl` is vanishing or something in the `#[Subscription]` macro, or its actually being packed up somewhere and the `dyn` is forgotten as it's now required and has been recommended for a couple years now.

1. Use the code from the examples in the docs:
    ```rs
    pub struct Subscription;
    
    #[Subscription]
    impl Subscription {
    	async fn interval(&self, #[graphql(default = 1)] n: i32) -> impl Stream<Item = i32> {
    		let mut value = 0;
    		async_stream::stream! {
    			loop {
    				tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
    				value += n;
    				yield value;
    			}
    		}
    	}
    }
    ```
2. Make sure your Cargo.toml is set to `edition = "2021"`
3. `cargo build`

And if you switch back to the 2018 edition then it works, then back to 2021 and it fails.

## Specifications

- Version:
  - `rg async-graphql Cargo.toml` -> `10:async-graphql = { version = "2.10.3", features = ["apollo_tracing", "uuid"] }`
  - `cargo version` -> `cargo 1.56.0 (4ed5d137b 2021-10-04)`
  - `rustc --version` -> `rustc 1.56.0 (09c42c458 2021-10-18)` Stable version
- Platform: Linux 64-bit
- Subsystem: `stable-x86_64-unknown-linux-gnu`

OvermindDL1:
Fixed #663 

This gets the `Subscription` procmacro working on Rust 2021, which is when the `dyn` keyword is now hard required.  The generated code was missing instances of the `dyn` keyword.

### DIFF
--- a/derive/src/subscription.rs
+++ b/derive/src/subscription.rs
@@ -219,7 +219,7 @@ pub fn generate(
                 let mut r = None;
                 for b in bounds {
                     if let TypeParamBound::Trait(b) = b {
-                        r = Some(quote! { #b });
+                        r = Some(quote! { dyn #b });
                     }
                 }
                 quote! { #r }
@@ -303,7 +303,7 @@ pub fn generate(
                         #(#schema_args)*
                         args
                     },
-                    ty: <<dyn #stream_ty as #crate_name::futures_util::stream::Stream>::Item as #crate_name::Type>::create_type_info(registry),
+                    ty: <<#stream_ty as #crate_name::futures_util::stream::Stream>::Item as #crate_name::Type>::create_type_info(registry),
                     deprecation: #field_deprecation,
                     cache_control: ::std::default::Default::default(),
                     external: false,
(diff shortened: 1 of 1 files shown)


OvermindDL1:
Think I found a fix, adding it to the main injected stream type, hold on...

OvermindDL1:
And all tests in my big project and in async-graphql are successful with `--all-features`, sec while I commit and update this...

OvermindDL1:
And done, this should be ready for testing/merge if someone else can confirm as well.  Tested on both 2018 and 2021.

sunli829:
Thanks! 😁

PR #667, opened 2021-10-22
feat: add derived field for object

Miaxos:
Add a `derived field` feature for the `Object` macro.

# Context

We have a usecase at work where we want to have multiple representation available for a Scalar or an Object: we want to have several formats possible for our business types.

To solve this issue, we imagined a way to generate associated fields based on a `From` implementation with the codegen. It's a little inspired by the derived field from the [Viaduct`s Airbnb talk](https://www.youtube.com/watch?v=xxk9MWCk7cM) where we could create with the codegen derived fields.

It's a first implementation which cover our usecase which is to create derived fields based on the `From` trait, but I would like to extend it to maybe support other things later.

In our usecase, the `ComplexObject` solution doesn't really fit, because we want to generate it inside a lot of objects and not just `SimpleObject`, so we tried to reduce the boilerplate.

# Example

```rust
#[derive(Enum, Eq, Clone, Copy, PartialEq, Hash, Debug)]
enum TestEnum {
  #[graphql(name = "MESSAGE")]
  Message = 0,
  #[graphql(name = "ENUM")]
  Enum = 1,
}

impl From<TestEnum> for u8 {
  fn from(value: TestEnum) -> u8 {
    match value {
      TestEnum::Message => 0,
      TestEnum::Enum => 1,
    }
  }
}

struct Query;

#[Object]
impl Query {
  #[graphql(derived(name = "test_u8", into = "u8"))]
  async fn test(&self, arg1: String) -> FieldResult<TestEnum> {
    println!("Test call");
    Ok(TestEnum::Message)
  }
}
```

It would generate an associated `test_u8` query to call and to return a Int instead of an Enum.

Would love to have your feedbacks!

### DIFF
--- a/docs/en/src/derived_fields.md
+++ b/docs/en/src/derived_fields.md
@@ -4,13 +4,13 @@ When you are working on a GraphQL project, you usually have to explain and share
 be interpreted by your consumers. Sometimes, you event want to have the same data and the same logic exposing
 the data in another type.
 
-Within `async-graphql` you can create derivated fields for objects to generate derivated fields.
+Within `async-graphql` you can create derived fields for objects to generate derived fields.
 
 Consider you want to create a `Date` scalar, to represent an event of time.
 How will you represent and format this date? You could create a scalar `Date` where you specified it's the RFCXXX
 implemented to format it.
 
-With derivated fields there is a simple way to support multiple representation of a `Date` easily:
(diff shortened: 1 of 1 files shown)


sunli829:
Cool, this is very useful!👏 I think `SimpleObject` macro and `ComplexObject` macro can also add this, what do you think?

PR #689, opened 2021-11-03
feat: add chrono::Duration custom scalar

meh:
For reference: https://www.graphql-scalars.dev/docs/scalars/duration

### DIFF
--- a/src/types/external/duration.rs
+++ b/src/types/external/duration.rs
@@ -1,0 +1,21 @@
+use chrono::Duration;
+use iso8601_duration as iso8601;
+
+use crate::{InputValueError, InputValueResult, Scalar, ScalarType, Value};
+
+/// Implement the Duration scalar
+///
+/// The input/output is a string in ISO8601 format.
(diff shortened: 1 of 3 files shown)


src/types/external/duration.rs
```diff
@@ -0,0 +1,21 @@
+use chrono::Duration;
+use iso8601_duration as iso8601;
+
+use crate::{InputValueError, InputValueResult, Scalar, ScalarType, Value};
+
+/// Implement the Duration scalar
+///
+/// The input/output is a string in ISO8601 format.
+#[Scalar(internal, name = "Duration")]
```
Miaxos, line 9: There is a new `specified_by_url` field specified by the Graphql 2021 spec to describe the current implementation:

```suggestion
#[Scalar(internal, name = "Duration", specified_by_url = "https://en.wikipedia.org/wiki/ISO_8601#Durations")]
```

meh:
Added that, should be good to go.

Miaxos (approved):

PR #692, opened 2021-11-04
Added map, contains and transpose function to MaybeUndefined

SadiinsoSnowfall:
This PR aims at improving the usability of MaybeUndefined: most operations should be possible without first converting the `MaybeUndefined<_>` into an `Option<Option<_>>`.

Exemple, decoding a base64 string given as an argument:
```rust
struct Changeset {
    bin: Option<Option<Vec<u8>>>,
}

// before
fn decode_before(param: MaybeUndefined<String>) -> Result<Changeset> {
    let decoded = Option::<Option<_>>::from(param).map(|v| v.map(base64::decode).transpose()).transpose()?;
    Ok(Changeset {
        bin: decoded
    })
}
// or
fn decode_before(param: MaybeUndefined<String>) -> Result<Changeset> {
    let decoded = match param {
        MaybeUndefined::Value(v) => Some(Some(base64::decode(v)?)),
        _ => None
    };
    
    Ok(Changeset {
        bin: decoded
    })
}

// after
fn decode_after(param: MaybeUndefined<String>) -> Result<Changeset> {
    let decoded = param.map_value(base64::decode).transpose()?;
    Ok(Changeset {
        bin: decoded.into()
    })
}
```

The name of the function might need to be changed. There is a need for two map/contains function because `MaybeUndefine` can have three states, as opposed to `Option` which only has two.

The `ok_*` and `map_*` functions from `Option` could also be ported but would be more complex to design, mainly because of the number of possible states.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `chrono::Duration` custom scalar. [#689](https://github.com/async-graphql/async-graphql/pull/689)
 - Implement `From<Option<Option<T>>>` for `MaybeUndefined<T>`.
-- Add `MaybeUndefined::as_opt_ref` and `MaybeUndefined::as_opt_deref` methods.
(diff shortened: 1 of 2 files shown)


sunli829:
It looks good, can you help update the update log? 🙂

SadiinsoSnowfall:
OK, I have updated the changelog to reflect the changes in this PR.

sunli829:
Thank you so much!😄

PR #695, opened 2021-11-09
Fix wrong result type in GraphQLSubscription

sarendsen:


sunli829:
`async-graphql-v3` has rework all the integrations, so this PR is no longer needed, But still thank you! 😁

PR #697, opened 2021-11-10
Async-graphql v3

sunli829:
This is a good opportunity. We can change the unreasonable design. If you have any ideas, please let me know. 😁

- [x] Use the `SimpleObject` macro and the `InputObject` macro at the same time.

```rust
#[graphql(input_name = "MyObjectInput")]
struct MyObject {
    #[graphql(default = 10)]
    a: i32,
    b: bool,
}
```

- [x] Types that are not referenced will be hidden in introspection.

```rust
#[derive(SimpleObject)]
struct MyObj {
    a: i32,
}

#[Object]
struct Query {
    #[graphql(visible = false)]
    fn obj(&self) -> MyObj {
        todo!()
    }
}

let schema = Schema::new(Query, EmptyMutation, EmptySubscription);
assert_eq!(
    schema.execute(r#"{ __type(name: "MyObj") { name } }"#).await.into_result().unwrap().data,
    value!({ "__type": null })
);
```

- [x] Make the API of integrations is more consistent.
- [X] Remove `async-graphql-tide`.
- [X] Rework validators.

Built-in validator:

```rust
struct Query;

#[Object]
impl Query {
    async fn value1(&self, #[graphql(validator(maximum = 10))] n: i32) -> i32 {
        n
    }
}
```

Custom-validator:

```rust
struct MyValidator {
    expect: i32,
}

impl MyValidator {
    pub fn new(n: i32) -> Self {
        MyValidator { expect: n }
    }
}

#[async_trait::async_trait]
impl CustomValidator<i32> for MyValidator {
    async fn check(&self, _ctx: &Context<'_>, value: &i32) -> Result<(), String> {
        if *value == self.expect {
            Ok(())
        } else {
            Err(format!("expect 100, actual {}", value))
        }
    }
}

struct Query;

#[Object]
impl Query {
    async fn value(
        &self,
        #[graphql(validator(custom = "MyValidator::new(100)"))] n: i32,
    ) -> i32 {
        n
    }
}
```

- [x] Rework guards.

```rust
#[derive(Eq, PartialEq, Copy, Clone)]
enum Role {
    Admin,
    Guest,
}

pub struct RoleGuard {
    role: Role,
}

impl RoleGuard {
    fn new(role: Role) -> Self {
        Self { role }
    }
}

#[async_trait::async_trait]
impl Guard for RoleGuard {
        if ctx.data_opt::<Role>() == Some(&self.role) {
            Ok(())
        } else {
            Err("Forbidden".into())
        }
    }
}

 #[derive(SimpleObject)]
struct Query {
    #[graphql(guard = "RoleGuard::new(Role::Admin)")]
    value: i32,
}
```

Use parameters:

```rust
struct EqGuard {
    expect: i32,
    actual: i32,
}

impl EqGuard {
    fn new(expect: i32, actual: i32) -> Self {
        Self { expect, actual }
    }
}

#[async_trait::async_trait]
impl Guard for EqGuard {
    async fn check(&self, _ctx: &Context<'_>) -> Result<()> {
        if self.expect != self.actual {
            Err("Forbidden".into())
        } else {
            Ok(())
        }
    }
}

struct Query;

#[Object]
impl Query {
    #[graphql(guard = "EqGuard::new(100, value)")]
    async fn get(&self, value: i32) -> i32 {
        value
    }
}
```

- [x] Update documents.


meh:
The only thing I need from the subscription stuff is to be able to extract authorization headers and such from the connection so I can authenticate the subscription.

leebenson:
> The only thing I need from the subscription stuff is to be able to extract authorization headers and such from the connection so I can authenticate the subscription.

It's generally a bad idea to rely on reading headers in WebSockets, since browsers have limited support for adding custom headers.

Instead, I'd rely on exchanging authorisation (or other) data on the initial payload.

meh:
You really don't want to have tokens accessible to the JS, you put them in HTTP only cookies.

leebenson:
> You really don't want to have tokens accessible to the JS, you put them in HTTP only cookies.

That won't work in all scenarios - e.g. if your GraphQL server (and therefore, WS(S) connection) is external to the mechanism that first retrieved and set the HTTP cookie.

sarendsen:
It would be nice if I could access the context in the middleware/connection_init 

right now I do authentication within them but as I cannot access the context I have to rely on global (database connection to verify the auth token)

meh:
>That won't work in all scenarios - e.g. if your GraphQL server (and therefore, WS(S) connection) is external to the mechanism that first retrieved and set the HTTP cookie.

That's nice and all, but that's not my situation, so I'd still like to be able to do it :)

leebenson:
> That's nice and all, but that's not my situation, so I'd still like to be able to do it :)

Understood. Just some advice for others reading this thread and trying to figure out auth -- a common pain point for GraphQL subscriptions, in my experience.

In either case -- here's [how gqlgen exposes the context to middleware](https://gqlgen.com/recipes/authentication/) (Go).

sunli829:
@meh I'm almost done, there will be everything you want. 🙂

sunli829:
I deleted the integration of `tide` because it does not support `websocket` very well.

sunli829:
@meh Here is an example of how to use request headers to create subscription context data. This example is based on `poem`. Now all web framework integration usage is similar.

Maybe you should try `poem`, I think it is very easy to use. 🙂

https://github.com/async-graphql/examples/blob/be1508f163311f0ca9ed987c9487fd7d28fac942/poem/token-from-header/src/main.rs#L44

meh:
@sunli829 is it me or there's no way to pass state to the `on_connection_init` callback?

I kind of really need that to fully support authentication through `GQL_CONNECTION_INIT`.

sunli829:
@meh I have full support for this in `Async-graphql v3` and also provide examples.

### Poem

https://github.com/async-graphql/examples/blob/ee6e8581f096001820c0d4829fa88c0b85f2f8c1/poem/token-from-header/src/main.rs#L44

### Actix-web

https://github.com/async-graphql/examples/blob/ee6e8581f096001820c0d4829fa88c0b85f2f8c1/actix-web/token-from-header/src/main.rs#L40

meh:
> @meh I have full support for this in `Async-graphql v3` and also provide examples.
> 
> ### Poem
> 
> https://github.com/async-graphql/examples/blob/ee6e8581f096001820c0d4829fa88c0b85f2f8c1/poem/token-from-header/src/main.rs#L44
> 
> ### Actix-web
> 
> https://github.com/async-graphql/examples/blob/ee6e8581f096001820c0d4829fa88c0b85f2f8c1/actix-web/token-from-header/src/main.rs#L40
> 

Yeah, but on_connection_init only takes the JSON value, and nothing else.

I need that function to get access to some additional state (database and such).

PR #707, opened 2021-11-15
Add unset method for ErrorExtensionValues

IGI-111:
I've been using this library to develop an API and I'm missing this particular small feature to allow me to clean up internal errors.

I can set the values to null as it is, but I'd rather not expose the existence of the field to an attacker altogether. And it doesn't seem like it would hurt anybody to have that feature.


sunli829:
Thank you, it has been manually merged into `async-graphql-v3`

PR #730, opened 2021-11-29
Add support for SmolStr via a feature.

lfn3:
SmolStr is an immutable "packed" string that uses the space normally occupied
by a String's header if the String is short enough.

I mostly copied the implementation from Uuid, which didn't appear to have any test coverage.
Happy to add some to this if you let me know roughly what the shape of the tests should be?


sunli829:
Thanks

PR #741, opened 2021-12-04
Fix implicit interfaces not being exposed via the __schema introspection

### LINKED ISSUE #740: Interfaces not being exported via schema in Async GraphQL 3.x
## Expected Behavior
registering an interface to properly expose it via the __schema API.

## Actual Behavior
The interface is not accessible via the __schema API.

1. I have two types that implement an interface
```rust
pub struct Question {
    pub video: Video,
}
#[Object]
impl Question {
    pub async fn id(&self) -> &VideoID {
        &self.video.id
    }
}

pub struct Response {
    pub video: Video,
}
#[Object]
impl Response {
    pub async fn id(&self) -> &VideoID {
        &self.video.id
    }
}
```
2. I have a union that that uses those 2 types
```rust
#[derive(Union)]
pub enum VideoUnion {
    Question(Question),
    Response(Response),
}
```
3. I have an interface registered via register_output_type.
```rust
#[derive(Interface)]
#[graphql(
    name = "Video",
    field(name = "id", type = "&VideoID"),
)]
pub enum VideoInterface {
    Question(Question),
    Response(Response),
}

// later in registering schema
schema.register_output_type::<VideoInterface>()
```

## Other information
1. the interface types DO get exposed via printing schema.sdl(), so it seems only that portion was perhaps updated with the split of registering output vs input types?
2. Note that all of my queries/endpoints return VideoUnion, Question, or Response and never "Video". 
3. this same setup works perfectly in 2.11.3

## Specifications

- Version: (only tested 3.0.7 and 3.0.11)
- Platform: macOS 10.15 / Rust 1.56
- Subsystem:

urkle:
This fixes implicit interfaces not being exposed in introspection as they used to in version 2.x.

Fixes #740

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -922,6 +922,19 @@ impl Registry {
             traverse_type(ctx, &self.types, &mut visible_types, ty.name());
         }
 
+        for ty in self.types.values() {
+            if let MetaType::Interface { possible_types, .. } = ty {
+                if ty.is_visible(ctx) && !visible_types.contains(ty.name()) {
+                    for type_name in possible_types.iter() {
+                        if visible_types.contains(type_name.as_str()) {
+                            traverse_type(ctx, &self.types, &mut visible_types, ty.name());
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
         self.types
             .values()
             .filter_map(|ty| {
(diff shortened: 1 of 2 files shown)


sunli829:
Thank you very much, this bug is hard to find. 🙂

PR #755, opened 2021-12-17
Fix compilation with change from poem 1.2.x

bradms:
This fixes the issue when compiling with `poem` 1.2.x where `Error::new_with_status` was renamed to `Error::from_status`.

Previously, it was failing with 
```
error[E0599]: no function or associated item named `new_with_status` found for struct `poem::Error` in the current scope
  --> integrations/poem/src/subscription.rs:33:35
   |
33 |             .ok_or_else(|| Error::new_with_status(StatusCode::BAD_REQUEST))
   |                                   ^^^^^^^^^^^^^^^ function or associated item not found in `poem::Error`
```


sunli829:
Thanks! 😁

PR #761, opened 2021-12-21
Improve regex validator error message to be descriptive

### LINKED ISSUE #760: Regex validator reports on error: value is valid - probably needs a better description
Subj https://github.com/async-graphql/async-graphql/blob/e0980106d8c5c2b6c4f01c6b513ab5f065a77f70/src/validators/regex.rs#L12

I'm not sure what exactly it supposed to mean but it is definitely not conveying that value validation failed due to not matching expected format by regex.

I think it would be better to modify it to something like:
`value doesn't match expected format`

What do you think?

DoumanAsh:
Fixes #760

### DIFF
--- a/src/validators/regex.rs
+++ b/src/validators/regex.rs
@@ -9,7 +9,7 @@ pub fn regex<T: AsRef<str> + InputType>(
     if let Ok(true) = Regex::new(regex).map(|re| re.is_match(value.as_ref())) {
         Ok(())
     } else {
-        Err("value is valid".into())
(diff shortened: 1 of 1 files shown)


sunli829:
Thanks!  🙂

DoumanAsh:
@sunli829 do you think it would be ok to include regex inside message?

sunli829:
Sounds good!👍

sunli829:
Thanks! 🙂

PR #765, opened 2021-12-22
fix: issue when empty with federation spec on _Entity node

Miaxos:
Hello 👋 

There is an issue with the current implementation of the `_Entity` node for federation.

[Reference](https://www.apollographql.com/docs/federation/federation-spec/#resolve-requests-for-entities)

**If no types are annotated with the key directive, then the _Entity union and Query._entities field should be removed from the schema.**


sunli829:
Thanks you so much! 🙂👏

PR #768, opened 2021-12-28
Add hashbrown HashMap/HashSet implementations for Input/Output type.

BratSinot:
Greetings!

I don't know if that make any sense or don't look like third wheel, but `hashbrow` one in one replacement for `HashMap` from std and I using it in my project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ async-stream = "0.3.0"
 async-trait = "0.1.48"
 fnv = "1.0.7"
 futures-util = { version = "0.3.0", default-features = false, features = ["io", "sink"] }
+hashbrown = { version="0.11.2", optional = true }
 indexmap = "1.6.2"
 once_cell = "1.7.2"
 pin-project-lite = "0.2.6"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -33,6 +33,7 @@ async-stream = "0.3.0"
 async-trait = "0.1.48"
 fnv = "1.0.7"
 futures-util = { version = "0.3.0", default-features = false, features = ["io", "sink"] }
+hashbrown = { version="0.11.2", optional = true }
```
sunli829, line 36: Move this line after `# Feature optional dependencies`.

sunli829 (commented):
Thank you very much! 🙂

It would be nice if you could help to add `hashbrown` feature to the `Crate Features` paragraph in `lib.rs` and `README.md`, and update `CHANGLOG`.

Cargo.toml
BratSinot: Also fixed rustfmt error (always forget to enable in my IDE for new project =) ).

sunli829:
Thanks!

PR #771, opened 2021-12-30
feat: added min_password_strength validator

oscartbeaumont:
This will still have to be added to the documentation but I can do that in a future PR.


sunli829:
Thanks!

PR #784, opened 2022-01-16
Actix integration: cbor response support + error handling improvements

Follpvosten:
This PR does two things:
- Remove `.unwrap()` from the `Responder` impl for GraphQLResponse, instead opting to act the same way Actix's `Json` responder does it: https://docs.rs/actix-web/4.0.0-beta.20/src/actix_web/types/json.rs.html#119-137
- Add cbor support to the same impl, reading the `Accept` header and acting based on the mime type in there.

I'm not 100% sure if this PR is in a mergable state, I could feature-flag the whole cbor thing or split this up into two PRs if that's preferred.

I've also noticed that the cbor request implementation doesn't actually support the `application/cbor` mime type atm, the stated reason being that `mime` doesn't support it yet; however `mime` is unmaintained and will never add it, and also it's possible to match on `(mime::APPLICATION, "cbor")` without the crate needing to add support. That's definitely a separate PR however (and I don't need cbor requests, so I'm not sure if I'm the right person to implement that).

### DIFF
--- a/integrations/actix-web/Cargo.toml
+++ b/integrations/actix-web/Cargo.toml
@@ -21,6 +21,7 @@ actix-web-actors = "4.0.0-beta.9"
 async-channel = "1.6.1"
 futures-util = { version = "0.3.0", default-features = false }
 serde_json = "1.0.64"
+serde_cbor = "0.11"
 serde_urlencoded = "0.7.0"
 futures-channel = "0.3.13"
 thiserror = "1.0.30"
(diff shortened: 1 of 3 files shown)


sunli829:
Thank you! It would be nice to have `serde_cbor` as an optional feature. 🙂

Follpvosten:
Alright, a couple notes:
* While I'm much more satisfied with this version, there might be some things still missing or could be improved
* I've basically copied the add test and adjusted it to request cbor. Because serde_cbor doesn't have a macro like serde_json's `json!`, I've derived Deserialize on a struct for that.

sunli829:
Nice, I'll merge it first and then refactor if necessary. I would like to move the cbor response to `async-graphql` so that all integrations can be easily supported.

PR #786, opened 2022-01-18
Add DataContext trait

smmoosavi:
Define the `DataContext` trait and implement for `ContextBase` and `ExtensionContext`

## Why

Sometimes we have a function that gets `Context` as an argument, we can't use that function in `Extension`

```rust
pub fn get_foo_services(ctx: &async_graphql::Context<'_>) -> impl FooServices {}

#[async_trait::async_trait]
impl Extension for FooExtension {
    async fn execute(
        &self,
        ctx: &ExtensionContext<'_>,
        operation_name: Option<&str>,
        next: NextExecute<'_>,
    ) -> Response {
        let result = next.run(ctx, operation_name).await;
        let services = get_foo_services(ctx);

        result
    }
}
```

```
         let services = get_hsm_services(ctx);
                                         ^^^ expected struct `async_graphql::ContextBase`, found struct `async_graphql::extensions::ExtensionContext`
```

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -22,6 +22,28 @@ use crate::{
     ServerResult, UploadValue, Value,
 };
 
+/// Data related functions of the context.
+pub trait DataContext<'a> {
+    /// Gets the global data defined in the `Context` or `Schema`.
+    ///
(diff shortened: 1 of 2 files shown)


sunli829:
This is very useful, thank you! 🙂

PR #795, opened 2022-01-24
Fixed variable type not checked when given a default value

SadiinsoSnowfall:
Given the following setup : 
```rust
 #[Object]
impl Query {
    pub async fn int_val(&self, value: Option<i32>) -> i32 {
        value.unwrap_or(10)
    }
}
```
This Request will fails and return the error message `Unknown Type "Invalid"` : 
```gql
query QueryWithVariables($intVal: invalid) {
    intVal(value: $intVal)
}
```
Whereas this one will cause a panic : 
```gql
query QueryWithVariables($intVal: invalid = 2) {
    intVal(value: $intVal)
}
```
This is because the code path for checking a variable with a default value doesn't check for the variable type. (see [this](https://github.com/async-graphql/async-graphql/blob/master/src/validation/rules/default_values_of_correct_type.rs#L21) and [this](https://github.com/async-graphql/async-graphql/blob/master/src/validation/utils.rs#L73)).

This PR adds two unit test to ensure that the problem is fixed and attempts to solve the problem by checking the variable type before checking the variable value.

### DIFF
--- a/src/validation/rules/default_values_of_correct_type.rs
+++ b/src/validation/rules/default_values_of_correct_type.rs
@@ -1,3 +1,5 @@
+use async_graphql_parser::types::BaseType;
+
 use crate::context::QueryPathNode;
 use crate::parser::types::VariableDefinition;
 use crate::validation::utils::is_valid_input_value;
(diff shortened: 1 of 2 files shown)


sunli829:
Thank you so much! 👏🏻

PR #817, opened 2022-02-10
Introduce process_with for input object

DoumanAsh:
This PR introduces attribute `process_with` which allows to specify path to a function that would accept mutable reference to the parsed field within `InputObject`

This is only done when object is supplied by user (e.g. default value is not post-processed) before validators are invoked

@sunli829 Let me know what you think. I found myself a need for this mostly because we need empty strings as a way to delete value(because we use `None` as indication of no change) and it is much better to safeguard myself from whitespace inputs by having this post-processor instead of rejecting input in validator, or doing post-processing yourself

### DIFF
--- a/derive/src/args.rs
+++ b/derive/src/args.rs
@@ -154,6 +154,8 @@ pub struct SimpleObjectField {
     pub visible: Option<Visible>,
     #[darling(default, multiple)]
     pub derived: Vec<DerivedField>,
+    #[darling(default)]
+    pub process_with: Option<String>,
     // for InputObject
     #[darling(default)]
     pub default: Option<DefaultValue>,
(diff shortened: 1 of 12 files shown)


DoumanAsh:
@sunli829 kindly ping, do you want it a good addition or?

DoumanAsh:
@sunli829 Sorry, but are you interested in this PR or not? If not I'll just close it

sunli829:
Sorry, I forgot about this. 😂

Currently the `process_with` attribute only supports input objects, it would be better to add all input value types.

-  Object field parameters
- Subscription field parameters
- Complex object field parameters

Use it like this:

```rust
struct Query;

#[Object]
impl Query {
    async fn value(&self, #[graphql(process_with="str::make_ascii_uppercase")] input: String) {
        ...
    }
}
````

What do you think? 😁

DoumanAsh:
Sounds good, I'll see how to do it

DoumanAsh:
@sunli829 because of fields being inherently more complex (namely user can specify mutability manually) I have to work it around by re-binding the same variable twice.
Resulting code is probably going to be optimized but it is still going to introduce some extra complexity

sunli829:
looks very good, thanks! 😁

DoumanAsh:
Nice, I'll be waiting for axum upgrade and then enjoy this feature in our code base

PR #831, opened 2022-02-18
update chrono-tz to 0.6.1

urkle:
Update dependency

Question..
Should we expose a feature to allow to still use the older chrono-tz? (Unsure how this manifests itself if the main app is trying to use the old chrono-tz)


sunli829:
Thanks! 😁

Issue #196, opened 2020-07-01
ScalarType for i/u64 represented as integer instead of string


danielSanchezQ:
In previous versions the `ScalarType` trait relied in a `to_json` method. It was possible to represent an int (i/u64) as a integer instead of a string for responses.
Right now the `ScalarType` relies on a `to_value` method in where any integer can be represented as a number just up to i32 (`Value(i32)`).  Completely invalidating the other possibility. 
Maybe can be a good idea to keep the `to_json` method with a default implementation. This way the standard can be kept as `Value(i32)` but it will still be possible to override that behavior for other purpouses.

This implementation is not possible anymore:
```rust
use async_graphql::{InputValueError, InputValueResult, ScalarType, Value};

// Wrapper around `i64` to force graphql to format it as a number (JSON) instead of a string
pub struct ScalarI64(pub i64);

#[async_graphql::Scalar]
impl ScalarType for ScalarI64 {
    fn parse(value: Value) -> InputValueResult<Self> {
        match value {
            Value::Int(n) => Ok(ScalarI64(n)),
            Value::String(s) => Ok(ScalarI64(s.parse()?)),
            _ => Err(InputValueError::ExpectedType(value)),
        }
    }

    fn to_json(&self) -> Result<serde_json::Value, async_graphql::Error> {
        Ok(serde_json::Value::Number(self.0.into()))
    }
}
```

A tiny example of troublesome is that if someone want to keep the responses from some REST endpoint and GraphQL offering the same results for an object, lets say: `struct Foo { bar : i64 }`:
The response object would look like,
REST:
```json
{
    "bar" : 1000000
}
```
GraphQL
```json
{
    "bar" : "1000000"
}
```

And the deserialization can be troublesome.

I'm down to try it and open a PR if it is something that can be considered.

Thanks

phated:
I'm pretty sure the JSON spec doesn't support BigInts, so I think you have a bug in your API

danielSanchezQ:
Maybe I was unfortunate. I meant big ints (i64, u64), not really BigInt. 
Referencing the JSON limit there is an interesting [post](https://stackoverflow.com/questions/13502398/json-integers-limit-on-size). 
I don’t think is really a bug since it was working as expected with plain json_serde (and it was working as expected before the change in the ScalarType trait). But I’m completely open to suggestions.

danielSanchezQ:
[...]

PR #844, opened 2022-02-23
custom_scalars: i64 not encoded as string

kolloch:
Remove the note that incorrectly claims that i64 are encoded as strings.

Also see https://github.com/async-graphql/async-graphql/issues/196


kolloch:
It is probably hard to change now without breaking compatibility but I find the default somewhat unsafe because graphql really only supports i32 as number.

https://spec.graphql.org/June2018/#sec-Int

PR #853, opened 2022-03-02
Correct import path for cursor connection in docs

Zwo1in:
As in title, just a little correction of paths for the book


sunli829:
Thanks! 😁

PR #856, opened 2022-03-04
Update actix to 0.13

skos-ninja:
Building the `actix-web` integration throws an error:
```
the trait `actix::actor::Actor` is not implemented for `subscription::GraphQLSubscriptionActor<Query, Mutation, Subscription, OnInit>`
```

Updating to actix 0.13 fixes this


sunli829:
Thanks! 🙂

skos-ninja:
Sorry to be a pain but would it be possible to get his out asap @sunli829 as `v3.0.32` failed to build based off of this error so the latest version is still using the pinned rc which isn't compatible with the final release 😢

PR #862, opened 2022-03-10
Update regex

Dylan-DPC:
Updates regex to `1.5.5` as per the [security advisory](https://blog.rust-lang.org/2022/03/08/cve-2022-24713.html)


sunli829:
Thanks! 🙂

PR #863, opened 2022-03-11
Make HashMap more generic.

BratSinot:
Greetings!

This adds state to `HashMap` generic to use different hashes (like for example `ahash`).


sunli829:
Thanks！😄

PR #875, opened 2022-03-23
Add feature(bson-uuid) which will enable Uuid's from the bson crate

serates:
Simple change but has great impact when using async-graphql with a mongodb database. Adds a new feature bson-uuid which should not be used in combo with uuid (due to them defining the same scalar differently). But this adds support for bson::Uuid which ensures that bson's are stored in the database using the proper structure while returning them through graphql api as strings.


sunli829:
Thanks! 😁

Issue #864, opened 2022-03-11
Treat &Type and Type in edge/connection as the same thing


Sytten:
## Description of the feature

Currently if you try to use a ref sometimes and the owned object in another place, you will get the error:
```
thread 'main' panicked at '`async_graphql::types::connection::edge::Edge<proxy_backend::api::Cursor, proxy_backend::api::SitemapEntry, async_graphql::types::connection::EmptyFields>` and `async_graphql::types::connection::edge::Edge<proxy_backend::api::Cursor, &proxy_backend::api::SitemapEntry, async_graphql::types::connection::EmptyFields>` have the same GraphQL name `SitemapEntryEdge`', /home/sytten/.cargo/registry/src/github.com-1ecc6299db9ec823/async-graphql-3.0.34/src/registry/mod.rs:411:25
```

This error makes sense when the type are different, but for a ref not so much.
This forces you to use `Cow` everywhere and it complexifies your code for no real benefit.

Sytten:
Otherwise maybe just add a feature to disable the panic?

sunli829:
This is really cumbersome and I haven't figured out a good way to fix this, adding a flag to turn off this detection looks nice. 😁

Sytten:
Fixed with the ignore name conflict

Issue #694, opened 2021-11-05
Handle possible errors related to MergedObject at compile time


SadiinsoSnowfall:
There is two kind of errors linked to `MergedObjects` which can compile but will cause a panic / unwanted behaviour at runtime.

#### Using an `EmptyMutation` in a `MergedObject` after another mutation
Example:
```rust
 #[derive(Default)]
pub struct Mutation;

#[Object]
impl Mutation {
    async fn test(&self) -> bool {
        true
    }
}

#[derive(MergedObject, Default)]
pub struct MutationRoot(EmptyMutation, EmptyMutation);  // is ok

#[derive(MergedObject, Default)]
pub struct MutationRoot(EmptyMutation, Mutation);  // is ok

#[derive(MergedObject, Default)]
pub struct MutationRoot(Mutation, EmptyMutation); // will panic at runtime

// using a schema built using a QueryRoot, mutation set to MutationRoot and subscription set to EmptySubscription
```

Using an `EmptyMutation` in a `MergedObject` at a later position compared to the mutation which will be executed during a request will result in a panic with the error `entered unreachable code` linked to this line [types/empty_mutation.rs](https://github.com/async-graphql/async-graphql/blob/master/src/types/empty_mutation.rs#L59); which executes the `unreachable!()` macro.

In this example, the server receive a query calling the `test` mutation.

#### Using multiple mutations with the same name in a `MergedObject`
Example:
```rust
pub mod mod1 {
    #[derive(Default)]
    pub struct Mutation;
}

pub mod mod2 {
    #[derive(Default)]
    pub struct Mutation;
}

#[derive(MergedObject, Default)]
pub struct MutationRoot(mod1::Mutation, mod2::Mutation);
```
If multiple mutation objects with the same name are attached to a `MergedObject`, only the first one will be resolved (in this example, the one from `mod1` although the program does compiles.

I think the `MergedObject` macros should prevent this example from compiling, to avoid weird behaviour at runtime.

SadiinsoSnowfall:
It should be noted that while the second problem is documented, or at least mentioned in the book :
> Tip: Every #[Object] needs a unique name, even in a MergedObject, so make sure to give each object you're merging its own name.
I think it would be best to prevent code that has this problem from compiling.

sunli829:
> It should be noted that while the second problem is documented, or at least mentioned in the book :
> 
[...]

PR #876, opened 2022-03-23
Control name conflict panic with feature

Sytten:
Fixes https://github.com/async-graphql/async-graphql/issues/864

I had to go the `disallow` route with default because all the tests run with `--all-features`.
Probably doesn't matter since users didn't disable default in the past considering it was empty.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["network-programming", "asynchronous"]
 readme = "README.md"
 
 [features]
-default = []
+default = ["disallow-name-conflict"]
 apollo_tracing = ["chrono"]
 apollo_persisted_queries = ["lru", "sha2"]
 unblock = ["blocking"]
@@ -24,6 +24,7 @@ decimal = ["rust_decimal"]
 cbor = ["serde_cbor"]
 chrono-duration = ["chrono", "iso8601-duration"]
 password-strength-validator = ["zxcvbn"]
+disallow-name-conflict = []
 
 [dependencies]
 async-graphql-derive = { path = "derive", version = "3.0.36" }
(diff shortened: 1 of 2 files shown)


sunli829:
Give me some time, I hope to find a better solution. 😁

sunli829:
> Its a tricky problem. Did you try it with stripped symbols to see if it still works correctly? The documentation kinda hints that it is a debug thing (aka not expected to be used in production).

Turning off duplicate name detection will cause serious problems, currently only `Connection` cannot be renamed, I'm still trying to figure out a way to fix it.

sunli829:
A temporary solution is to define your own `XXXConnection` type instead of the generic `Connection` provided in `async-graphql`.

Sytten:
Yeah we are going that direction its just that we have a lot of places to change and I have to rewrite everything (`query` function, `cursor`, `edge`, etc). So It's fine if you don't want that merged, I will use our own fork until I move to our own connections.

SadiinsoSnowfall:
See #864 and #694 (which is the issue that lead to adding the "panic on shared name" check).

Sytten:
Any progress/decision on what you want to do moving forward? @sunli829

sunli829:
Allowing this feature to be turned off causes inexplicable errors, and I really haven't figured out what to do. 😂

Sytten:
It used to work well in v2 FYI since we abuse that feature right now. Worst case I will use Cow everywhere, just annoying. No worries :)

sunli829:
Sorry to keep you waiting, I'm working on a fix for this in `async-graphql-v4`, basically adding a generic parameter to the `Connection` and `Edge` types for a custom name. 😁

Sytten:
In my case I dont believe that will help since it is a matter of using a ref vs an owned object, but a good thing anyway!
I started migrating to Cow object to fiz the issue.

Sytten:
Hey @sunli829, what is your take moving forward? Using the name trait to differentiate?
It is not a fix for me since I will still need to start using Cow everywhere to support both refs and owned objects but its probably fair to do so.

sunli829:
I'll add a option tomorrow that allows you to specify which names are allowed to be excluded from detection.

PR #884, opened 2022-04-06
Support macro type in enum variant

cynecx:
This is useful when union types (eg. result types) are generated by some `macro_rule`s.


sunli829:
Thanks! 😁

PR #885, opened 2022-04-07
Added instructions for returning errors from subscription resolvers

### LINKED ISSUE #845: Returning from subscription with error
It could happen that a subscription fails and instead of returning something like 

```
impl Stream<Item = Option<MyThing>>
```

I would like to interrupt the stream or not start it at all.
How can this be done with the current implementation?

Venryx:
^Title.

The example code in the linked comment is fairly long, so I just linked to it rather than including it in the page (also because my solution for changing the error-type within the resolver is probably more verbose than needed).

But the new explanation on the page at least sets people on the correct path.

Resolves #845 (mostly anyway).

### DIFF
--- a/docs/en/src/error_handling.md
+++ b/docs/en/src/error_handling.md
@@ -25,3 +25,12 @@ impl Query {
     }
 }
 ```
+
+#### Errors in subscriptions
+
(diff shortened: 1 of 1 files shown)


sunli829:
Thanks! 🙂

PR #894, opened 2022-04-18
Introspection-only mode

cynecx:
This is useful for "proxies" (Hasura's remote schema feature) that only need access to introspection.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -6,6 +6,7 @@ use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::parser::parse_query;
 use crate::parser::types::ExecutableDocument;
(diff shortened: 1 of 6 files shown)


src/request.rs
```diff
@@ -14,6 +15,7 @@ use crate::{Data, ParseRequestError, ServerError, UploadValue, Value, Variables}
 /// variables. The names are all in `camelCase` (e.g. `operationName`).
 #[derive(Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
+#[non_exhaustive]
```
cynecx, line 18: This is actually a breaking change (semver). But `https://github.com/async-graphql/async-graphql/pull/891` has already introduced such thing....

PR #896, opened 2022-04-19
Implemented OutputType for tokio RwLock

boardmaster357:


sunli829:
What do you think if it would be better to implement both for `tokio::sync::Mutex`? 😁

boardmaster357:
Implementing for Mutex as well would be a good idea. Would you like me to revise the pull request? Should we just call the compile option "tokio"?

sunli829:
> Implementing for Mutex as well would be a good idea. Would you like me to revise the pull request? Should we just call the compile option "tokio"?

Yes, and how about we named the feature `tokio-sync`? 😁

boardmaster357:
Great idea. Will do.

boardmaster357:
Thanks for the great improvemnts. I updated the pull request.

sunli829:
Thank you so much! 😁

PR #901, opened 2022-04-25
parser: move code generation into a test

djc:
Move the pest parser code generation step into a test. Makes it so that downstream users don't have to spend the time compiling some of the dependencies (in this case, pest_meta and its sha-1 dependency).

Inspired by https://github.com/steveklabnik/semver-parser/pull/50. It's not as big an improvement here because we still depend on serde's `Serialize` and `Deserialize` macros, but seems like a nice improvement anyway.

### DIFF
--- a/parser/Cargo.toml
+++ b/parser/Cargo.toml
@@ -14,6 +14,9 @@ categories = ["network-programming", "asynchronous"]
 [dependencies]
 async-graphql-value = { path = "../value", version = "3.0.38" }
 pest = "2.1.3"
-pest_derive = "2.1.0"
 serde_json = "1.0.64"
 serde = { version = "1.0.125", features = ["derive"] }
+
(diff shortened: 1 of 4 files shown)


sunli829:
LGTM 🙂

sunli829:
@djc An absolute path will be generated here, do you know how to solve it? 😁

https://github.com/async-graphql/async-graphql/blob/ad05baf79aea55a0827291abf55af5ce7cc8d386/parser/src/parse/generated.rs#L8

djc:
Ah, so I guess the pest derive macro expands the relative path to the grammar to an absolute path? I guess you can just patch up the generated code in the code generation test, by generating the same absolute path and replacing it with a relative path?

Also, what's the reason for merging this into the v4 branch? It seems semver-compatible, so could just go in v3, right?

Issue #916, opened 2022-05-06
Implement OneofObjectType for Vec<T>


bvanneerven:
## Description of the feature

Implementing `OneofObjectType` for `Vec<T>` would allow a user of the library to use a list of some input object type as an argument.

## Code example (if possible)

```rs
use async_graphql::*;

#[derive(SimpleObject)]
pub struct User {
    name: String,
}

#[derive(OneofObject)]
pub enum UserBy {
    Email(String),
    RegistrationNumber(i64),
}

pub struct Query {}

#[Object]
impl Query {
    #[graphql(oneof)] // causes error
    async fn search_users(&self, by: Vec<UserBy>) -> Vec<User> {
        unimplemented!()
    }
}
```
```gql
directive @oneOf on INPUT_OBJECT

type User {
  name: String!
}

input UserBy @oneOf {
  email: String
  registrationNumber: Int
}

type Query {
  searchUsers(by: [UserBy!]!): [User!]!
}
```

Removing `#[graphql(oneof)]` from the code above actually compiles and throws an error when more than one field is defined on the `UserBy` input object at runtime. However, right now, removing this also means that the directive is not shown in the output GraphQL schema.

sunli829:
I added a test and everything seems to be working fine.

https://github.com/async-graphql/async-graphql/blob/336451f3bc5d422dc9c6ab9744c728984bedfbce/tests/oneof_object.rs#L268

bvanneerven:
Jup, that test should indeed work. I tried to describe this in the issue, but maybe I was not clear enough. 
The issue is that the GraphQL schema now misses the declaration of the directive:
```gql
directive @oneOf on INPUT_OBJECT
```
As per https://github.com/async-graphql/async-graphql/issues/766#issuecomment-1059242265 we should include the `#[graphql(oneof)]` macro to include this.
If you include that macro above the line in the test here you should receive an error. https://github.com/async-graphql/async-graphql/blob/336451f3bc5d422dc9c6ab9744c728984bedfbce/tests/oneof_object.rs#L286

![Cargo check error](https://user-images.githubusercontent.com/35584266/167250106-b0030fe3-67db-462f-8058-3c74e49c5c45.png)

Hope this helps with understanding this issue a bit more ☺️

sunli829:
I have removed the `oneOf` attribute, it is no longer needed and it looks like the generated SDL is correct now. 🙂

```graphql
directive @oneOf on INPUT_OBJECT

type Query {
[...]

PR #917, opened 2022-05-07
Add OneofObject to book

bvanneerven:
After our discussion in https://github.com/async-graphql/async-graphql/issues/916 I thought it would be nice to add the discussed example to the book, such that users of the library might discover the feature more quickly ☺️


sunli829:
👏Thank you so much!

PR #926, opened 2022-05-20
fix #925 support BigDecimal and support parse from Number for Decimal

### LINKED ISSUE #925: Support BigDecimal
Hi, `BigDecimal` seems does not support by this crate. Could this type be supported? It is used in `sqlx` and other ORM liberary.

al8n:
Hi, this PR supports BigDecimal and supports parse from Number for Decimal.


sunli829:
Thank you so much! 🙂

PR #931, opened 2022-05-27
Expose IntrospectionMode

DoumanAsh:
While field was added to `Request`, type itself is hidden so I think would be good to expose it?


sunli829:
Thanks! 🙂

PR #939, opened 2022-05-31
update darling version

urkle:
- zxcvbn uses this newer version as well

- according to the changelog for darling we **should** be able to remove the `#[darling(default)]` from any of the Option<> fields.  Let me know if you want me to do that for this PR.


urkle:
@sunli829 I rebased this against current master and all tests are passing.

PR #940, opened 2022-06-02
run book tests and fix book examples

urkle:
run mdbook tests so we ensure the examples actually compile

There are several examples that were broken and did not actually work (thus making it hard to learn to use async-graphql)


sunli829:
awesome, thank you so much for making this! 👏

PR #950, opened 2022-06-14
Make email-validator optional

### LINKED ISSUE #949: Make `fast_chemail` dependency optional
version: v4.0.2

Hi! We're running into licensing issues using this library unfortunately due to the inclusion of an MPL-2.0 licensed dependency

```sh
$ cargo-license
...
MPL-2.0 (2): ascii_utils, fast_chemail
```
Its brought in by the `fast-chemail` library used for email validation - (it in turn drags in `ascii_utils`).

In our case, we don't have a need for email validator but would still like to use the library so would you folks be open to making `fast_chemail` an optional dependency so that we can opt out of it? (If so we'd be happy to create a PR for this)

iffy-validio:
This change makes it possible to opt out of the email validator
which brings in the `fast_chemail` crate. Since it won't be
ideal to break existing usages, by default the dependency is
still included.

Fixes #949

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,11 +15,12 @@ version = "4.0.2"
 [features]
 apollo_persisted_queries = ["lru", "sha2"]
 apollo_tracing = ["chrono"]
+email-validator = ["fast_chemail"]
 cbor = ["serde_cbor"]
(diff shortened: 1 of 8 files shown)


sunli829:
Thanks 😊

PR #955, opened 2022-06-20
Add a prefer_single_line_descriptions option on SDLExportOptions

ivan:
The default export uses three lines for each description. Single-line descriptions can improve the readability of the exported SDL when there are many short descriptions.

With `prefer_single_line_descriptions `, the exported output looks like
https://gist.github.com/ivan/f8e394651a86a016a7d99258fbc556ff
instead of
https://gist.github.com/ivan/9b0e473e4a2a600d93f1b1610a40260a

Please let me know what you think, I am happy to change things if you want.


sunli829:
Thanks! 🙂

PR #969, opened 2022-06-30
Fix serializing of JSON default values

urkle:
This changes the InputType::to_value implementations to serialize the JSON values to strings. (since these are scalar types GraphQL requires it to be a string for the default)


sunli829:
Thanks! 🙂

Issue #900, opened 2022-04-20
SimpleObject: error: lifetime may not live long enough


bbigras:
I got some weird error after using `#[graphql(complex)]` and `#[ComplexObject]` on other struct.

So the affected struct doesn't have the "complex" thing.

I'm probably using nested `ComplexObject` that uses some dataloaders.

I'll try to make a minimal, reproducible example.

## Actual Behavior

```
13 | #[derive(Debug, PartialEq, SimpleObject)]
   |                            -^^^^^^^^^^^
   |                            |
   |                            let's call the lifetime of this reference `'2`
   |                            let's call the lifetime of this reference `'1`
   |                            associated function was supposed to return data with lifetime `'1` but it is returning data with lifetime `'2`
   |                            in this derive macro expansion
   |
  ::: /home/bbigras/.cargo/registry/src/github.com-1ecc6299db9ec823/async-graphql-derive-3.0.38/src/lib.rs:47:1
```

## Steps to Reproduce the Problem

1.
2.
3.

## Specifications

- Version: 3.0.38
- Platform: NixOS 22.05 (Quokka) x86_64
- Subsystem:

sunli829:
Please provide an example that reproduces this error.

kennetpostigo:
I get the same thing locally since i've updated async-graphql, it happens sporadically

kennetpostigo:
After a clean build it goes away but comes back after a while

bbigras:
> After a clean build it goes away but comes back after a while

I had something similar too.

sunli829:
Could you please provide an example that can reproduce this problem?😁

bbigras:
Yes sorry. I'll be back home tomorrow. (I should have still made one earlier as promised when I opened the issue though)

bbigras:
I think I'm able to reproduce. There's a compiler bug, but I'm not sure yet if both problems are related.

```
error: lifetime may not live long enough
  --> src/query.rs:53:26
   |
53 | #[derive(Clone, Default, SimpleObject)]
   |                          -^^^^^^^^^^^
   |                          |
   |                          let's call the lifetime of this reference `'1`
   |                          let's call the lifetime of this reference `'2`
   |                          associated function was supposed to return data with lifetime `'2` but it is returning data with lifetime `'1`
   |
   = note: this error originates in the derive macro `SimpleObject` (in Nightly builds, run with -Z macro-backtrace for more info)

[...]

PR #972, opened 2022-07-04
Palliative workaround for #900

danielhenrymantilla:
This uses <https://docs.rs/fix-hidden-lifetime-bug> to replace the compiler's heuristic with lifetimes with a simpler and more straight-forward one, which **does dodge the regression from https://github.com/rust-lang/rust/issues/98890**.

# How to use

Add the following to your workspace[^1]'s `Cargo.toml` file:

```toml
[patch.crates-io.async-graphql]
git = "https://github.com/danielhenrymantilla/async-graphql.git"
branch = "workaround-for-1-62-bug"
# or, instead of branch, you may pin-point its `HEAD` commit, if you are scared of a sneaky branch update
# rev = "6dae5c864c009de9dda03fc0b7796ced02e895bd"
```

[^1]: the directory where the `Cargo.lock` file is generated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ async-graphql-value = { path = "value", version = "4.0.4" }
 async-stream = "0.3.0"
 async-trait = "0.1.48"
 bytes = { version = "1.0.1", features = ["serde"] }
+fix-hidden-lifetime-bug = "0.2.5"  # See #900
 fnv = "1.0.7"
 futures-util = { version = "0.3.0", default-features = false, features = [
   "io",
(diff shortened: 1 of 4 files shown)


sunli829:
Thanks a lot, it's strange that I didn't have this problem. 😂

ControlCplusControlV:
@danielhenrymantilla looks like you may need a `cargo fmt --all`. I ran into this issue as well and ended up here so I hope you get your PR merged soon!

ivan:
I confirmed that with this PR applied to async-graphql, I cannot reproduce the issue in my comment https://github.com/async-graphql/async-graphql/issues/900#issuecomment-1175787840

leefordjudes:
this is not worked for my team after adding this patch., so, we reverted to rust 1.61.

danielhenrymantilla:
@leefordjudes interesting, could you share what errors you got with this patch?

danielhenrymantilla:
No longer needed as of the new toolchain release

PR #983, opened 2022-07-14
Add generics support for Interface and Union static assertions

### LINKED ISSUE #978: Generics in resolvers for Interface
Hello, I'm trying to create an interface where resolvers are generic for a type `T`, because I want to get from the context that specific type. This process works well with `Object`s but seems to break with an `Interface`.
Here's the example code to reproduce it:

```rust
use async_graphql::*;
use async_trait::async_trait;
use std::marker::PhantomData;

#[derive(Interface)]
#[graphql(
    field(name = "field", type = "String"),
)]
pub enum Multiple<T: SomeTrait> {
    One(One<T>),
    Two(Two<T>)
}

pub struct One<R>(PhantomData<R>, String);
#[Object]
impl<R: SomeTrait> One<R> {
    async fn field<'a>(&self, ctx: &Context<'a>) -> String {
        let _a = ctx.data::<R>().unwrap();
        format!("one: {}", self.1)
    }
}

pub struct Two<R>(PhantomData<R>, String);
#[Object]
impl<R: SomeTrait> Two<R> {
    async fn field<'a>(&self, ctx: &Context<'a>) -> String {
        let _a = ctx.data::<R>().unwrap();
        format!("two: {}", self.1)
    }
}

#[async_trait]
pub trait SomeTrait: 'static + Send + Sync {}

```
But when I try to compile the above I get:
```
error[E0412]: cannot find type `T` in this scope
  --> src/lib.rs:10:13
   |
10 |     One(One<T>),
   |             ^ not found in this scope

error[E0412]: cannot find type `T` in this scope
  --> src/lib.rs:11:13
   |
11 |     Two(Two<T>)
   |             ^ not found in this scope

For more information about this error, try `rustc --explain E0412`.
error: could not compile `async-graphql-example2` due to 2 previous errors
```

Do you have some suggestions on how to get the above to work?

Using `cargo expand` the issue seems to come from this const function block:
```
const _: fn() =
    ||
        {
            use ::static_assertions::_core::marker::PhantomData;
            use ::static_assertions::_core::ops::Deref;
            let previous = AssertImplAnyFallback;
            struct AssertImplAnyFallback;
            struct ActualAssertImplAnyToken;
            trait AssertImplAnyToken {}
            impl AssertImplAnyToken for ActualAssertImplAnyToken {}
            fn assert_impl_any_token<T: AssertImplAnyToken>(_: T) {}
            let previous =
                {
                    struct Wrapper<T, N>(PhantomData<T>, N);
                    impl <T, N> Deref for Wrapper<T, N> {
                        type Target = N;
                        fn deref(&self) -> &Self::Target { &self.1 }
                    }
                    impl <T: async_graphql::ObjectType, N> Wrapper<T, N> {
                        fn _static_assertions_impl_any(&self)
                         -> ActualAssertImplAnyToken {
                            ActualAssertImplAnyToken
                        }
                    }
                    Wrapper::<One<TRAITOR>, _>(PhantomData, previous)
                };
            let previous =
                {
                    struct Wrapper<T, N>(PhantomData<T>, N);
                    impl <T, N> Deref for Wrapper<T, N> {
                        type Target = N;
                        fn deref(&self) -> &Self::Target { &self.1 }
                    }
                    impl <T: async_graphql::InterfaceType, N> Wrapper<T, N> {
                        fn _static_assertions_impl_any(&self)
                         -> ActualAssertImplAnyToken {
                            ActualAssertImplAnyToken
                        }
                    }
                    Wrapper::<One<TRAITOR>, _>(PhantomData, previous)
                };
            assert_impl_any_token(previous._static_assertions_impl_any());
        };
```

`T` has been renamed to `TRAITOR`, but anyway that `Wrapper` tries to use the `TRAITOR` type which is not in scope.

Thanks for the library and any help is appreciated!

choco:
This is an attempt to fix #978. This unfortunately relies on an unreleased version of `static_assertions` crate

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ pin-project-lite = "0.2.6"
 regex = "1.5.5"
 serde = { version = "1.0.125", features = ["derive"] }
 serde_json = "1.0.64"
-static_assertions = "1.1.0"
+static_assertions = { git = "https://github.com/nvzqz/static-assertions-rs", rev = "a867e97a1b3c2e0c76b7a154abdf4e422b78e5f2" }
 tempfile = "3.2.0"
 thiserror = "1.0.24"
 
--- a/derive/src/interface.rs
+++ b/derive/src/interface.rs
@@ -18,6 +18,7 @@ use crate::{
 pub fn generate(interface_args: &args::Interface) -> GeneratorResult<TokenStream> {
     let crate_name = get_crate_name(interface_args.internal);
     let ident = &interface_args.ident;
+    let type_params = interface_args.generics.type_params().collect::<Vec<_>>();
     let (impl_generics, ty_generics, where_clause) = interface_args.generics.split_for_impl();
     let s = match &interface_args.data {
         Data::Enum(s) => s,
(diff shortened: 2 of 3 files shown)


choco:
Hey @sunli829, sorry for the ping. Is this a use case that you're interested in supporting? (in a codebase I work on we use this, but if there's no plan to support generics I may look into migrating to trait objects)

TroyKomodo:
@sunli829 this pr can be closed after the merge of #1424

choco:
ty @TroyKomodo

PR #995, opened 2022-07-26
Implement Content-Encoding aware query decompression support

### LINKED ISSUE #879: Take the content-encoding header into account when parsing a query
Currently, async-graphql reads the `content-type` header when parsing a GQL query to know if the query is encoded in JSON or CBOR.
However, it doesn't read the `content-encoding` header, which might indicate that the query was compressed.

Async-graphql should read the `content-encoding` header and decompress the payload if needed before parsing the query according to the `content-type` header. I think it would be sufficient for a first implementation of this feature to only handle `gzip`  compression; and this could be guarded behind a crate feature to mirror what was done with CBOR decoding.

Request compression is a niche use case but it can substantially reduce the volume of data sent to the GQL server. This is why I think it could be important to support this.

SadiinsoSnowfall:
Just a draft PR, this needs verification & unit testing.

This PR implements Content-Encoding aware decompression support for incoming GQL queries. This can be useful when dealing with large queries that compress well. This add supports for gzip, deflate, brotli and zstd compression algorithms under the new feature "compression".

Actix-web has similar features to allow the decompression of incoming requests based on the content-encoding header.

Fixes #879.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ string_number = []
 tokio-sync = ["tokio"]
 tracing = ["tracinglib", "tracing-futures"]
 unblock = ["blocking"]
+compression = ["brotli", "flate2", "zstd"]
 
 [dependencies]
 async-graphql-derive = { path = "derive", version = "4.0.6" }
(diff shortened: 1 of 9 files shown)


sunli829:
Thanks a lot for the PR, just this should be implemented by web framework? 😂

SadiinsoSnowfall:
This is indeed already implemented by some web frameworks (such as Actix-web) but async-graphql bypasses this mechanism because it works on raw requests.

PR #996, opened 2022-07-28
Add tracing to dataloader methods when the tracing feature is enabled.

### LINKED ISSUE #908: Tracing with Dataloader does not work correctly
## Expected Behavior
I would expect the spans that are created within the dataloader to have a parent span that is generated by the tracing extension

## Actual Behavior
When using an tracing extension with async-graphql and dataloaders the traces get split up and the parent-child relationship gets lost.

Enable the tracing extension and then load multiple items using a dataloader. Notice that all spans that are generated within the dataloader are not connected to the parent graphql span 

## Specifications

- Version: 3.0.31
- Platform: Linux
- Subsystem: tracing

Dig-Doug:
I tested this by modifying the tide-dataloader example and attaching it to jaeger. You'll see in the screenshot below that the trace successfully gets passed from the top level schema, through the library, and finally to the data loader implementation.

localhost_16686_trace_4baff6fd542886d9c118deb6bed16483 (2)

Fixes #908

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ tracing-futures = { version = "0.2.5", optional = true, features = [
   "std-future",
   "futures-03",
 ] }
-tracinglib = { version = "0.1.25", optional = true, package = "tracing" }
+tracinglib = { version = "0.1.35", optional = true, package = "tracing" }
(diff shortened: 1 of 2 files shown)


sunli829:
That's great! 🙂

PR #1010, opened 2022-08-05
Update README.md

bradleyess:
Add Formidable blog post for Async GraphQL w/ Rust. 

I found this to be an incredibly "real-world" example of using this library and I think lots of others will find similar value.


sunli829:
Thanks! 🙂

PR #1013, opened 2022-08-08
Add section about redacting data from input objects

Gricha:
Hey! I've spent a bit more time than necessary to find out about `secret` attribute. Throwing this into the book's InputObject section.


sunli829:
Thanks!😃

PR #1018, opened 2022-08-11
Fixes extension request.data(X) being lost in the resolver

k-kinzal:
The extension request.data(X) is lost in the resolver.
```rust
async fn prepare_request(
    &self,
    ctx: &ExtensionContext<'_>,
    request: Request,
    next: NextPrepareRequest<'_>,
) -> ServerResult<Request> {
    request.data("foo".to_string())
    next.run(ctx, request).await
}
```
```rust
#[Object]
impl Query {
    async fn resolve(&self, ctx: &Context<'_>) -> Result<Vec<Account>> {
        ctx.data_unchecked::<String>(); // panic occurs
}
```

I fixed it because I want to do JWT authentication in extension.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -254,6 +254,7 @@ pub struct QueryEnvInner {
     pub uploads: Vec<UploadValue>,
     pub session_data: Arc<Data>,
     pub ctx_data: Arc<Data>,
+    pub extension_data: Arc<Data>,
     pub http_headers: Mutex<HeaderMap>,
     pub introspection_mode: IntrospectionMode,
     pub errors: Mutex<Vec<ServerError>>,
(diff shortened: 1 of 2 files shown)


sunli829:
Thanks!  🙂

sunli829:
I didn't consider that you would add new data on `prepare_request`.

k-kinzal:
@sunli829 
Oh! I thought new data could be added because `request.data` can be called `prepare_request`.

It is different from other extensions in that `prepare_request` can modify the content of the request.
Therefore, shouldn't the modified request content have an effect?

sunli829:
You are right, this is useful! 🙂

PR #1036, opened 2022-08-22
export_sdl: avoid trailing space for scalar definitions

dspicher:
This has been introduced in 619013d.

The problem can easily be avoided by modifying the `write!`
directives inside the federation if-clause to write space
prefixes instead of suffixes.


dspicher:
rebased and adapted for the new federation `write!`s.

sunli829:
Thanks!❤️

Issue #982, opened 2022-07-14
Federation 2 supported?


hchockarprasad:
Do async-graphql currently support federation 2 schemas?? For using it with new apollo router built with rust instead of nodejs apollo gateway..

peacechen:
We're deciding between Rust and Go to implement GraphQL sub-graphs. The Apollo page lists Rust as having very little Fed2 support:
https://www.apollographql.com/docs/federation/supported-subgraphs/#rust

Will this library be updated regularly? Is it production ready? Are there other (preferably official) Rust GraphQL libraries?

sunli829:
I'll support it when I'm free. 🙂

peacechen:
Fair enough. What that means is this project isn't ready for production use.

leebenson:
> Will this library be updated regularly?

I suggest you take a look at the releases, commits and issue board. @sunli829 is generously donating an enormous amount of time to supporting this project, and I for one am very grateful.

> Is it production ready?

https://github.com/async-graphql/async-graphql#whos-using-async-graphql-in-production

> Fair enough. What that means is this project isn't ready for production use.

That's terribly unfair. Many companies and projects are using it successfully in production. If your particular use-case isn't answered, perhaps you could consider sponsoring the project or contributing to it?

raptros:
I am not very familiar with the async-graphql codebase, but I do want to contribute to the community in any way I can here. I think it'd be helpful if we at least have ideas written down about how to implement each of the remaining pieces.

looking over the apollo compat list, i think the things we need still are:
* `@link` directive
* `@tag` directive
* `@override` directive
* plus ftv1 might be a "nice-to-have"

the override directive should be pretty straightforward since it only applies to fields; the changes that have been made so far should provide a good example for how to proceed. 

I managed to find a definition of the tag directive - it's documented here: https://specs.apollo.dev/tag/v0.2/; i think it should be pretty similar to existing directive support as well.

ftv1 is almost straightforward as per the apollo documentation: https://www.apollographql.com/docs/federation/metrics/#how-tracing-data-is-exposed-from-a-subgraph
it just requires looking for a header in the http request, and then computing the value to put in the response extensions. does anyone have any thoughts about how to do this, given the various web frameworks async-graphql integrates with?

the really tricky one is `@link`. I think there are several challenges here based on what I've seen from Apollo about its use, which seem to require totally new features in async-graphql.
1. we need a way to have schema-level directives. 
[...]

PR #1038, opened 2022-08-22
Add Apollo Federation `@tag` directive support

nmoutschen:
This adds the Apollo Federation `@tag` directive support to all supported locations (FIELD_DEFINITION, INTERFACE, OBJECT, UNION, ARGUMENT_DEFINITION, SCALAR, ENUM, ENUM_VALUE, INPUT_OBJECT, INPUT_FIELD_DEFINITION) with this, the only piece missing for https://github.com/async-graphql/async-graphql/issues/982 is the `@link` directive at the top of the schema.


src/registry/export_sdl.rs
```diff
@@ -207,8 +217,13 @@ impl Registry {
                     }
                     write!(sdl, "scalar {} ", name).ok();
 
-                    if options.federation && *inaccessible {
-                        write!(sdl, "@inaccessible ").ok();
+                    if options.federation {
+                        if *inaccessible {
+                            write!(sdl, "@inaccessible ").ok();
```
raptros, line 222: btw this other MR https://github.com/async-graphql/async-graphql/pull/1036 wants to change this line (and the write a few lines above) to move the space to the front.

sunli829:
Thanks!❤️

Issue #1033, opened 2022-08-21
Opaque cursor


plippe:
Hi,

GraphQL.org suggests having opaque cursors for pagination. They encourage a base 64 encoding.
> As a reminder that the cursors are opaque and that their format should not be relied upon, we suggest base64 encoding them.
> \- https://graphql.org/learn/pagination/

I currently use the following for my pet project

```rust
pub struct Base64Cursor<A>(pub A);

impl<A: CursorType> CursorType for Base64Cursor<A> {
    type Error = String;

    fn decode_cursor(s: &str) -> Result<Self, Self::Error> {
        base64::decode_config(s, base64::URL_SAFE_NO_PAD)
            .ok()
            .as_ref()
            .and_then(|s| std::str::from_utf8(s).ok())
            .and_then(|c| A::decode_cursor(c).ok())
            .map(|a| Base64Cursor(a))
            .ok_or_else(|| "Unable to parse cursor".to_owned())
    }

    fn encode_cursor(&self) -> String {
        self.0
            .encode_cursor()
            .pipe(|cursor| base64::encode_config(cursor, base64::URL_SAFE_NO_PAD))
    }
}
```

My implementation builds on top of `CursorType`, but it could use Serde's `Serialize` and `Deserialize`. Possibly something like the following:

```rust
// WIP, won't compile
impl<A: Serialize + Deserialize> CursorType for A {
    type Error = ???;

    fn decode_cursor(s: &str) -> Result<Self, Self::Error> {
        base64::decode_config(s, base64::URL_SAFE_NO_PAD)
            .and_then(|s| serde_json::from_str(s))
    }

    fn encode_cursor(&self) -> String {
        serde_json::to_string(self)
            .and_then(|json| base64::encode_config(json, base64::URL_SAFE_NO_PAD))
    }
}
```

While this is one idea, another option is to use [cbor](https://cbor.io/) as it is already included in this project.

Would there be any objections if I opened a PR to change `CursorType`'s `encode_cursor` and `decode_cursor`? I am open to any encoding if base64 isn't preferred.

onx2:
I [posted](https://github.com/async-graphql/async-graphql/issues/563) about this a while back and one of the folks at Vector, @leebenson,  posted a link to their open source code that solved 99% of the problems I was trying to solve. Maybe this will help you?

https://github.com/vectordotdev/vector/blob/master/src/api/schema/relay.rs

[...]

PR #1040, opened 2022-08-23
Opaque cursor

plippe:
Change the CursorType to generate a cursor for any type that implements Serde's `Serialize` and `Deserialize`. This removes the need to define implementations for built-in types and offers support for complex types. 

This PR is in response to https://github.com/async-graphql/async-graphql/issues/1033

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ async-graphql-value = { path = "value", version = "4.0.11" }
 
 async-stream = "0.3.0"
 async-trait = "0.1.48"
+base64 = "0.13.0"
 bytes = { version = "1.0.1", features = ["serde"] }
 fnv = "1.0.7"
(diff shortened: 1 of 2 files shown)


sunli829:
Thank you very much for the PR, but I think this change loses flexibility, in fact we just need to add a new `OpaqueCursor` to achieve this. 🙂

plippe:
Yes, some flexibility is lost as every cursor would now be opaque. I will update the PR and add a OpaqueCursorType. 

Just as a heads up, this approach will require the cursor type to be explicite. This isn't the end of the world, but I thought it was worth mentioning.

plippe:
It seems this was released, https://github.com/async-graphql/async-graphql/commit/337279f6c76885e1f792cff50e3e455efd664812
Closing the PR

PR #1043, opened 2022-08-25
export_sdl: consistently avoid trailing spaces

### LINKED ISSUE #1037: 4.0.10 adds trailing spaces in schema output
After upgrading to 4.0.10, the schema output changed to include trailing spaces after input type fields. Would be nice to avoid this (and other unnecessary changes to schema output, ideally).

dspicher:
116f6e48 modified the SDL export for scalars not to
incur trailing spaces. Unfortunately this occurs for
other types as well. This commit takes care of the
remaining occurrences and adds assertions to unit-tests
where an SDL is already available.

Closes #1037.

### DIFF
--- a/src/registry/export_sdl.rs
+++ b/src/registry/export_sdl.rs
@@ -271,29 +271,29 @@ impl Registry {
                     write!(sdl, "extend ").ok();
                 }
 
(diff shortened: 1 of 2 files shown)


sunli829:
Thanks! 🙂

PR #1044, opened 2022-08-27
Add GraphiQL v2

bvanneerven:
GraphiQL version 2 was [released](https://github.com/graphql/graphiql/releases) this week! 🎉 
It is the successor to both GraphiQL v1 and the GraphQL Playground (see https://github.com/graphql/graphql-playground/issues/1143 for more info).

I therefore propose to add the GraphiQL version 2 source to the repo and created this PR for it.
I created a separate module for this source, since that ensures nothing breaks 🙃.

Since this officially replaces the earlier versions of GraphiQL and GraphQL Playground it might be good to add the `#[deprecated]` attribute to `http::graphiql_source::graphiql_source` and `http::playground_source::playground_source` since these packages likely won't receive any further updates, but I want to leave that decision to you.

When this gets merged, I will also create a PR to update the examples in the other repo.

### DIFF
--- a/src/http/graphiql_v2_source.rs
+++ b/src/http/graphiql_v2_source.rs
@@ -66,7 +66,7 @@ impl<'a> GraphiQLSource<'a> {
         };
 
         r#"
-      <!DOCTYPE html>
+        <!DOCTYPE html>
         <html>
           <head>
             <style>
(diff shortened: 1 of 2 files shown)


bvanneerven:
The failing check has the error: 
```
error: package `time v0.3.14` cannot be built because it requires rustc 1.59.0 or newer, while the currently active 
rustc version is 1.57.0
```
I shouldn't really do anything with that I suppose 🤷

sunli829:
Thanks!😀

PR #1048, opened 2022-08-29
Compare to expected schema

### LINKED ISSUE #1045: SDL: missing space after input type name before open brace (without annotations)
Old is expected, new is actual:

```diff
@@ -175,9 +175,9 @@ type DomainConnection {
        pageInfo: PageInfo!
 }
 
-input DomainContactInput {
-       domain: String! 
-       message: String! 
+input DomainContactInput{
+       domain: String!
+       message: String!
 }
 
 type DomainEdge {
```

I think this is a regression from #1043. cc @dspicher

Perhaps there should be some tests added to make sure regressions in the SDL output are caught in CI?

dspicher:
The first commit adds logic to two unit tests where the schema
export is compared against expected output cached in two
schema files.

This is intended to help prevent bugs like the ones fixed in
https://github.com/async-graphql/async-graphql/commit/116f6e48be245097d7134407f723ef64edc2a91f or the immediately succeeding commit, as unexpected
changes to schema export will become apparent in the commit
diff, easing reviews.

When legitimately changing the export output behaviour,
the test suite just needs to be run twice, as the unit-tests
automatically overwrite the files with the new version.

This unit-test approach is inspired by

https://matklad.github.io/2022/03/26/self-modifying-code.html

The second commit fixes #1045.

### DIFF
--- a/src/registry/export_sdl.rs
+++ b/src/registry/export_sdl.rs
@@ -367,7 +367,7 @@ impl Registry {
                         write!(sdl, " @tag(name: \"{}\")", tag.replace('"', "\\\"")).ok();
                     }
                 }
-                writeln!(sdl, "{{").ok();
+                writeln!(sdl, " {{").ok();
 
                 let mut values = enum_values.values().collect::<Vec<_>>();
                 if options.sorted_enum_values {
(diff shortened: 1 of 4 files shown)


sunli829:
Thanks! 🙂

PR #1049, opened 2022-08-30
Support for primitive type in CursorType

k-kinzal:
Changed `CursorType` to support primitive types.
This is useful if the use case is to decode recursively.


sunli829:
Thanks! 🙂

PR #1060, opened 2022-09-04
implement a simple approach to using the link directive

raptros:
Part of adding fed 2 support (see #982).

Since async-graphql does not presently have any mechanism to attach custom schema directives, there's no need to support renaming directives (since we know we won't have any name conflicts), nor any reason to support importing schema directives beyond the apollo ones (since they wouldn't be usable anyway).

In a future where custom schema directives are supported, this implementation will need to be replaced (presumably using the mechanisms that will support custom schema directives to configure the attached ink directives, plus some mechanism for handling renaming of "known" directives).

Todo for the current approach
- [x] write tests to verify the new printing
- [x] apply apollo's testing harness to verify that this approach is in fact sufficient
- [x] decide: should enable_apollo_link be processed to implicitly enable federation (like having entities does?)

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -413,6 +413,11 @@ pub struct MetaDirective {
     pub visible: Option<MetaVisibleFn>,
 }
 
+#[derive(Debug, Clone)]
+pub struct ApolloLinkConfig {
+    pub extend_schema: bool,
+}
+
 #[derive(Default)]
 pub struct Registry {
     pub types: BTreeMap<String, MetaType>,
@@ -423,6 +428,7 @@ pub struct Registry {
     pub subscription_type: Option<String>,
     pub introspection_mode: IntrospectionMode,
     pub enable_federation: bool,
+    pub enable_apollo_link: bool,
     pub federation_subscription: bool,
     pub ignore_name_conflicts: HashSet<String>,
 }
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -423,7 +423,7 @@ pub struct Registry {
     pub subscription_type: Option<String>,
     pub introspection_mode: IntrospectionMode,
     pub enable_federation: bool,
-    pub enable_apollo_link: bool,
+    pub suppress_apollo_link: bool,
     pub federation_subscription: bool,
     pub ignore_name_conflicts: HashSet<String>,
 }
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -413,11 +413,6 @@ pub struct MetaDirective {
     pub visible: Option<MetaVisibleFn>,
 }
 
-#[derive(Debug, Clone)]
-pub struct ApolloLinkConfig {
-    pub extend_schema: bool,
-}
-
 #[derive(Default)]
 pub struct Registry {
     pub types: BTreeMap<String, MetaType>,
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -423,7 +423,7 @@ pub struct Registry {
     pub subscription_type: Option<String>,
     pub introspection_mode: IntrospectionMode,
     pub enable_federation: bool,
-    pub suppress_apollo_link: bool,
+    pub enable_apollo_link: bool,
     pub federation_subscription: bool,
     pub ignore_name_conflicts: HashSet<String>,
 }
--- a/docs/en/src/apollo_federation.md
+++ b/docs/en/src/apollo_federation.md
@@ -20,6 +20,26 @@
 
 - The `override` directive is used to indicate that a field is now to be resolved by the current subgraph instead of the named subgraph.
 
+- The `link` directive is needed in order to indicate that the subgraph is Federation v2 compatible, enabling the `shareable`, `inaccessable`, and `override` directives.
+
+## Enabling Federation v2 using the link directive
+
+async-graphql provides a configuration function `enable_apollo_fed2_link` on the schema builder to have it print out an `extend schema` element with an appropriately configured `link` directive whenever the federation schema is requested.
+
+```rust
+Schema::build(Query, EmptyMutation, EmptySubscription)
+    .enable_apollo_fed2_link()
+    .finish()
+```
+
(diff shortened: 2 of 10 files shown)


dbanty:
This looks great, definitely the simplest way to add `@link` support quickly.

> apply apollo's testing harness to verify that this approach is in fact sufficient

I ran some tests from your branch, and it looks good! The `@link` checkbox in our docs will turn green with this enabled 😄.

> decide: should enable_apollo_link be processed to implicitly enable federation (like having entities does?)

Maybe rename this to `enable_federation_v2` and have it set the same option as `enable_federation`? That makes it clear that the router/gateway will operate with v2, which does not have the same default behaviors as v1.

dbanty:
Hey, @raptros is there anything I can do to help finish this off? I'm itching to start using Fed 2 features with `async-graphql` subgraphs 😁

raptros:
thank you for running the apollo tests! just checked off that item. 

I want to add a test that works similarly to https://github.com/async-graphql/async-graphql/blob/master/tests/federation.rs#L542 - where it basically verifies by file matching - should be the most effective way to ensure the scheme works as expected.

re changing the name ... on the one hand, i see your point, on the other hand it doesn't seem to me to match how other things in async-graphql are named (where it seems somewhat more explicit about what you're getting).

dbanty:
> re changing the name ... on the one hand, i see your point, on the other hand it doesn't seem to me to match how other things in async-graphql are named (where it seems somewhat more explicit about what you're getting).

I think it's in the same spirit as `enable_federation`, where calling that method is actually adding `_service` and `_entities` types to the `Query`. It abstracts away the details of _how_ federation works and just states _that_ it will work. Likewise, I don't think most developers need to know about `@link` to start running with federation v2.

That said, it's really up to @sunli829 to decide what makes the most sense for this project 😁.

sunli829:
@raptros Thank you so much for your PR!

Currently `enable_federation` does not need to be called manually by the user. If there are any entities in the generated schema, the Federation related type will be registered.

https://github.com/async-graphql/async-graphql/blob/0eb9cf12d2e1366373336f254711205d29fa3af5/src/schema.rs#L225

So the name `enable_apollo_fed2_link` might be better than `enable_federation_v2`, since it always has to be called when federation 2 support is needed, or renamed to `enable_apollo_link` to be more concise.

@dbanty What do you think?

raptros:
that's a good point sunli, i hadn't thought about it that way (that the enable federation call usually does not need to be made, but this method will for now have to be made).

i'm wondering therefore if the actual best name might be `print_apollo_link` to make it really clear what the user is opting-into. 

but really i don't have any strong opinions on the name, let me know if you want me to change it (or leave it as-is).

dariuszkuc:
Presence of `@link` is used to determine whether it is federation v2 compatible schema or not. If you only want to support v2 (which eventually should be true) then IMHO there is no need for extra property and `@link` should be added whenever entities are detected.

If you want to support both v1 (current) and v2 (new) then maybe better word is "opt-in" - users have to explicitly opt into specific behavior (whether it is opt in v2 or downgrade to v1 is up to you).

sunli829:
I think @dariuszkuc is right，we can remove this option.😄

raptros:
changed it up to default to attaching it, and switched around the flag to make it suppress the printing.

I kept the option attached to the schema registry because users who want to suppress the directive will probably want to suppress it whenever the schema is printed in response to a federator's introspection request.

dbanty:
I agree that default would be the best experience for most users. My only concern is that it's potentially a breaking change for folks already using these subgraphs.

If someone ran an `async-graphql` subgraph within a fed2 supergraph and then updated to this new behavior, the composition may break. Specifically, all their types/fields would have been treated as `@shareable` before but no longer are once `@link` is added. See the note within [these docs](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#shareable) for more info.

Maybe this would affect few enough people that it's not a concern here, but I wanted to make sure it was brought up before this is merged 😁.

raptros:
@dbanty as i understand it, once a federator has updated the supergraph to fed2, new versions of subgraphs have to opt into fed2 in order to be published.

dbanty:
> @dbanty as i understand it, once a federator has updated the supergraph to fed2, new versions of subgraphs have to opt into fed2 in order to be published.
> 
> ie a user would have to opt-in to the fed2 semantics in order to deploy the new version anyway.

I'm fairly certain that isn't the case. At least with [router](https://www.apollographql.com/docs/router/), you can continue to run and publish new versions of fed1 subgraphs even though the supergraph is fed2. The whole idea is that fed2 should be adopted incrementally with minimal disruption. Maybe this is a limitation of certain implementations of gateways? I pretty much only use router 😅

sunli829:
@raptros I think we still keep `enable_apollo_fed2_link` and tell users in the docs (https://github.com/async-graphql/async-graphql/blob/master/docs/en/src/apollo_federation.md) to enable Federation 2 support It needs to be called manually.

raptros:
Sounds good to me. Restored the opt-in method, and wrote up some documentation.

src/registry/mod.rs
```diff
@@ -423,7 +423,7 @@ pub struct Registry {
     pub subscription_type: Option<String>,
     pub introspection_mode: IntrospectionMode,
     pub enable_federation: bool,
-    pub suppress_apollo_link: bool,
+    pub enable_apollo_link: bool,
```
dariuszkuc, line 426: Just wondering, instead of `enable_apollo_link` maybe better name would be `enable_federation_v2` or something along those lines. Outside of the library maintainers I am unsure whether devs would be aware of the significance here.

sunli829:
I decided to merge it first and then release the `5.0` version.

raptros:
Thank you Sunli!

sunli829:
`Async-graphql v5.0` will only support `Federation 2`, compatibility with `Federation 1` does not seem to make much sense. 😂

sunli829:
@raptros sorry I removed some of your code. 😂

dbanty:
Awesome, thanks @sunli829 !!!

PR #1062, opened 2022-09-05
Resurrect code generation through tests

rsdy:
This is a port of @djc's original PR #901.

`pest_generate` [allows disabling](https://docs.rs/pest_generator/latest/pest_generator/fn.derive_parser.html) the inclusion of grammar as part of a parameter, so I believe this should also work on the CI, as there are no absolute paths involved in the artifact.


sunli829:
Haha, Thanks!

PR #1066, opened 2022-09-11
Create a BaseLoader which can be given a non-static lifetime.

Wopple:
The motivation here is testing. I want to be able to create different instances of the same loader with different lifetimes. I need a static lifetime in production so the loader can cache values and be shared across threads. I need a loader with scope limited lifetimes in test scenarios so it can contain non-static entities.

I created a new trait called `BaseLoader` which can have any lifetime. `Loader` now is a sub-trait of `BaseLoader` and forces the lifetime to be static and forces the keys, values, and errors to be static as well (in addition to the other constraints). I also created some convenience traits to reduce some of the verbosity.

### DIFF
--- a/src/dataloader/cache.rs
+++ b/src/dataloader/cache.rs
@@ -5,24 +5,30 @@ use std::{
     marker::PhantomData,
 };
 
+/// Convenience trait with the constraints required for a cache key.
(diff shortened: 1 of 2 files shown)


Wopple:
Closing for now, this doesn't actually work yet...

PR #1080, opened 2022-09-22
docs: Update federation docs with examples of each directive

dbanty:
This greatly expands the federation docs by adding examples of all the directives as well as links out to the Apollo docs for each of them.

I may have gone a little overboard with the amount of examples 😅 but I wanted it to be very easy for folks to find and add a federated directive when they need one 😁.

I didn't add all of the `# ...` lines to make the code compilable, so I'm guessing CI will fail. If it does, I'll go back and add those soon!


sunli829:
Thanks a lot! 🙂

dbanty:
I think I fixed all the issues in the examples so they pass tests now! 🎉. Only remaining CI failures should be fixed by #1082

PR #1083, opened 2022-09-23
Add ignore_introspection to opentelemetry extension

onx2:
This seems to be a point of frustration for a few folks as the introspection polling causes a very noisy trace log. Going to open this up to start a discussion on if this should be added and what you think the proper way to do it is.

Refs:
https://github.com/async-graphql/async-graphql/issues/841


sunli829:
Added in 13a66a50

sunli829:
Added in 13a66a5013f05a691c5d042850548fde9f941f0a

onx2:
Thank you!

PR #1091, opened 2022-09-29
Use `tracinglib::error! instead of `tracinglib::info!`

frederikhors:
On [Rust forum](https://users.rust-lang.org/t/is-tracing-intended-to-be-used-as-i-am-using-it-how-can-i-trace-errors-chain/81988/2) they're helping me better understand https://github.com/async-graphql/async-graphql/discussions/1089 and found that we're calling

```rust
tracinglib::info!
```

instead of

```rust
tracinglib::error!
```

here: https://github.com/async-graphql/async-graphql/blob/974f8ff21d9019bde138f11c4eb5b1aeae554aa0/src/extensions/tracing.rs#L146

Is this wrong, right?

### DIFF
--- a/src/extensions/tracing.rs
+++ b/src/extensions/tracing.rs
@@ -144,7 +144,7 @@ impl Extension for TracingExtension {
         next.run(ctx, info)
             .map_err(|err| {
                 tracinglib::error!(target: "async_graphql::graphql",
-                                  error = %err.message,
+                                  error = tracinglib::field::display(&err.message),
                                   "error");
                 err
             })
--- a/src/extensions/tracing.rs
+++ b/src/extensions/tracing.rs
@@ -147,9 +147,9 @@ impl Extension for TracingExtension {
         };
 
         let fut = next.run(ctx, info).inspect_err(|err| {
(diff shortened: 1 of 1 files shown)


frederikhors:
Also here: https://github.com/tokio-rs/tracing/discussions/2329 they suggest the `info!` issue and more (I'll open an issue to discuss further the rest).

sunli829:
Maybe we should make the error log level to an option.

frederikhors:
> Maybe we should make the error log level to an option.

What do you mean?

sunli829:
Add a `Tracing::with_level` method to set the level of error log, what do you think?

frederikhors:
> Add a `Tracing::with_level` method to set the level of error log, what do you think?

I do not understand. I'm using `Tracing::with_level` in my app already, globally.

In this code I think we are only using the `error` level to trace an error bubbled up in our resolvers...

Am I wrong?

If I'm wrong please edit my PR: no problem at all! 😃

frederikhors:
@sunli829 can you please merge this and release?

PR #1105, opened 2022-10-09
Allow to add fetch `credentials` option to graphiql 2

it-nalon:
Added the fetch `credentials` option to the `GraphiQLSource` builder.

Example usage:
```rust
GraphiQLSource::build()
  .endpoint("http://localhost:8000")
  .credentials("omit")
  .finish();
```


sunli829:
Thanks! 🙂

PR #1110, opened 2022-10-11
Add TypeName support for InputObject

karatakis:
Add support for type_name derive argument for InputObject.

This will allow me to implement the trait from a custom derive I created, for more info look [here](https://github.com/SeaQL/seaography/blob/e7fe48798c589a42ca718dab63938f2a3333087d/derive/src/filter.rs#L107).


sunli829:
Thanks a lot, this is useful, I'll add this to other types as well.

PR #1119, opened 2022-10-24
Add support for 'input_name_suffix_ in '#[graphql(...)]' (fixes #1117)

### LINKED ISSUE #1117: Automatically generate an input name with a suffix
Hi!

Would it be possible, when using both `SimpleObject` and `InputObject` on the same struct, or both `Union` and `OneofObject`, to make the input type automatically suffixed?

For instance, the following code:

```rust
#[graphql(input_name_suffix = "Input")]
struct Test { /* ... */ }
```

Would now compile just like we wrote:

```rust
#[graphql(input_name = "TestInput")]
struct Test { /* ... */ }
```

This would be immensely helpful in macros, especially as `#[graphql(input_name = "...")]` doesn't accept variables.

What do you think of this?

ClementNerma:
Support for `SimpleObject`, `InputObject`, `Union` and `Oneofobject`.

Also improves refactoring:

```rust
// This:
#[graphql(input_name = "TestInput")]
struct Test { /* ... */ }

// Becomes:
#[graphql(input_name_suffix = "Input")]
struct Test { /* ... */ }
```

So when `Test` is renamed, the input name will be changed automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       matrix:
         include:
           - { rust: stable, os: ubuntu-20.04 }
-          - { rust: 1.64.0, os: ubuntu-20.04 }
+          - { rust: 1.65.0, os: ubuntu-20.04 }
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -44,7 +44,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { rust: 1.64.0, os: ubuntu-20.04 }
+          - { rust: 1.65.0, os: ubuntu-20.04 }
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Install stable toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.64.0
+          toolchain: 1.65.0
           override: true
       - name: Run cargo-tarpaulin
         uses: actions-rs/tarpaulin@v0.1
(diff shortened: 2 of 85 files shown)


sunli829:
Thanks for the PR, but I don't think the `input_name_suffix` attribute simplifies anything, so I won't merge this. 🙂

ClementNerma:
> Thanks for the PR, but I don't think the `input_name_suffix` attribute simplifies anything, so I won't merge this. 🙂

I strongly disagree, it enables several things:

1. You don't have to play with concatenation in macros
2. This avoids relying on an external dependency just for concatenation
3. When renaming a struct, the input type will be renamed automatically as well

I think this simplifies things a lot and I'd even argue that having this argument is more important than `input_name` itself as it serves more use cases :)

teenjuna:
@sunli829 are you sure? I can see why someone may want this, and the added code is quite simple. Maybe reconsider?

bvanneerven:
> 3. When renaming a struct, the input type will be renamed automatically as well

I like the change mainly because of this. 

When both `input_name` and `input_name_suffix` are used at the same time (for some reason?), I think `input_name` should override or an error should be thrown though.

ClementNerma:
> > 3. When renaming a struct, the input type will be renamed automatically as well
> 
> I like the change mainly because of this.
> 
> When both `input_name` and `input_name_suffix` are used at the same time (for some reason?), I think `input_name` should override or an error should be thrown though.

Not necessarily, for instance you could want to give a suffix to all of your types even with a custom name.

ClementNerma:
@sunli829 hi, I've rebased the PR to make it work with the latest 5.x versions. May you please reconsider the purpose of this add? I think it would be really useful for some people (like me) and it doesn't hurt to add it in the end for people that don't want to use it.

In my current project I constantly use my fork to have this feature which is essential.

PR #1131, opened 2022-11-03
Include enum value descriptions in sdl 

w0rm:
Hello and thanks for the great package!

I noticed that the generated sdl did not contain descriptions of enum values. This PR should fix this.


sunli829:
Thanks!😄

PR #1132, opened 2022-11-05
Make multipart a feature, and tempfile an optional dep

alexanderkjeldaas:
tempfile needs to generate random file names, which brings in crypto dependencies.
This is not always a requirement in constrained environments like in a WASM-build.


sunli829:
Thanks a lot, since this is a breaking change, I will merge it in `v5.0`.

sunli829:
I added the `tempfile` feature in `v5.0.0-alpha1`, so this PR is no longer needed, but still thanks for your help. 🙂

PR #1141, opened 2022-11-10
Fix regression on `ComplexObject` descriptions

95ulisse:
Hello :wave: 

The latest changes on dynamic schema introduced a regression on the `ComplexObject` macro where having a description set on any field would cause the build to break due to the generated code missing a call to `ToString::to_string()`.


sunli829:
Thanks!😄

PR #1142, opened 2022-11-10
Keep object 'implements' order stable in SDL export

sandhose:
In a project, I have the generated schema commited in the repository to help seeing diffs in the generated schema.

Problem is, the SDL output is not stable: the interface list order in a type `implements` changes on each run, leading to diffs like that:

```diff
- type Foo implements CreationDate & Node {
+ type Foo implements Node & CreationDate {
```

This is because internally it uses a `HashSet`. This changes PR that to an `IndexSet` to keep the insertion order.


sunli829:
Thanks!😄

PR #1153, opened 2022-11-28
Add boolean dynamic ValueAccessor method

### LINKED ISSUE #1152: Dynamic ValueAccessor needs a bool method
## Description of the feature
https://github.com/async-graphql/async-graphql/blob/bc4dfb503f36ca5a3d244b0346afaf0268e13a26/src/dynamic/value_accessor.rs

currently, Value Accessor has a large number of accessors, but bool values are excluded

## Code example (if possible)
```


    /// Returns the bool value
    pub fn bool(&self) -> Result<bool> {
        if let Value::Boolean(value) = self.0 {
            return Ok(*value);
        }
        Err(Error::new("internal: not a bool"))
    }
```

spiegela:
Adds a method to return a boolean from a `ValueAccessor`. Resolves #1152 1152

### DIFF
--- a/src/dynamic/value_accessor.rs
+++ b/src/dynamic/value_accessor.rs
@@ -15,6 +15,14 @@ impl<'a> ValueAccessor<'a> {
         matches!(self.0, Value::Null)
     }
 
+    /// Returns the boolean
+    pub fn boolean(&self) -> Result<bool> {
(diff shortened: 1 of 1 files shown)


sunli829:
Thanks! 🙂

PR #1182, opened 2022-12-22
Improve GraphiQL v2

bvanneerven:
Hi!

I am requesting some improvements to the GraphiQL v2 source generator.

**What has changed?**
- The source generator now uses a templating solution (handlebars), which makes the generator more robust. We can directly pass the `GraphiQLSource` struct to generate the source, instead of doing a lot of 'manual' string based manipulations. This introduces a dependency, but I think it is worth it from a long-term maintenance perspective.
- Changed `credentials` field to an enum. I think that is better than allowing arbitrary strings. **[breaking change]**
- The `endpoint` and `subscription_endpoint` are now relative urls instead of absolute ones. This matches with how routes are defined in most web frameworks; `Route::new().at("/", ...)` in poem for example. This way, we do not have to know the exact origin url when developing the application, increasing the 'seperation of concerns'. **[breaking change]**

I also created a PR to update the examples: https://github.com/async-graphql/examples/pull/64.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,6 +104,7 @@ lru = { version = "0.7.1", optional = true }
 serde_cbor = { version = "0.11.1", optional = true }
 sha2 = { version = "0.10.2", optional = true }
 zxcvbn = { version = "2.1.2", optional = true }
+handlebars = "4.3.6"
 
 [dev-dependencies]
 futures-channel = "0.3.13"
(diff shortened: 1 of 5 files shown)


sunli829:
Thanks! 👏

floric:
Hey, I just noticed the new dependency (handlebars) from 5.0.5. As I never use GraphiQL, can we make this optional as a feature flag to keep the release smaller? If there is nothing against it, I could create a MR. :)

sunli829:
Yes,welcome to PR😄

floric:
> Yes,welcome to PR😄

ok, will give it a try :)

floric:
> Yes,welcome to PR😄

see https://github.com/async-graphql/async-graphql/pull/1202 :)

I wasn't sure if we should set it as a default feature or use opt-in as done in this PR. But I assume, users will be experienced enough to enable the feature if needed. Most likely most will stick to the ready to use integrations and there it's not used right now. Should be still a minor change though.

Issue #1184, opened 2022-12-28
async-graphql-axum doesn't build with target wasm32-unknown-unknown


bbigras:
## Actual Behavior

I got `error[E0432]: unresolved import `crate::sys::IoSourceState`.

I wonder if it's because that tokio is not optional.

Note that `async-graphql` does build with target wasm32-unknown-unknown.

## Steps to Reproduce the Problem

1.
2.
3.

## Specifications

- Version:
- Platform:
- Subsystem:

bbigras:
I'm able to build with:

```diff
diff --git a/integrations/axum/Cargo.toml b/integrations/axum/Cargo.toml
index b8573b06..36e28c45 100644
--- a/integrations/axum/Cargo.toml
+++ b/integrations/axum/Cargo.toml
@@ -15,10 +15,10 @@ version = "5.0.4"
 async-graphql = { path = "../..", version = "5.0.4", default-features = false }

 async-trait = "0.1.51"
-axum = { version = "0.6.0", features = ["ws", "headers"] }
+axum = { version = "0.6.0", default-features = false, features = ["headers"] }
 bytes = "1.0.1"
 futures-util = "0.3.0"
 http-body = "0.4.2"
 serde_json = "1.0.66"
-tokio-util = { version = "0.7.1", features = ["io", "compat"] }
+tokio-util = { version = "0.7.1", default-features = false, features = ["io", "compat"] }
 tower-service = "0.3"
diff --git a/integrations/axum/src/lib.rs b/integrations/axum/src/lib.rs
index db48c948..e7cdbedc 100644
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -4,8 +4,8 @@

 mod extract;
 mod response;
-mod subscription;
+// mod subscription;

 pub use extract::{GraphQLBatchRequest, GraphQLRequest};
 pub use response::GraphQLResponse;
-pub use subscription::{GraphQLProtocol, GraphQLSubscription, GraphQLWebSocket};
+// pub use subscription::{GraphQLProtocol, GraphQLSubscription, GraphQLWebSocket};
```

PR #1185, opened 2022-12-28
integrations/axum: tokio: default-features = false

bbigras:
for wasm, see #1184

It doesn't fix everything, since I still need to remove the "ws" feature from axum, to build, but it's a start.


sunli829:
I guess this will fail to compile, we should add a `ws` feature to `async-graphql-axum`.

PR #1188, opened 2022-12-29
Fix: `__Type.oneOf` to `__Type.isOneOf`

bvanneerven:
Renamed the `@oneOf` directive introspection output according to the current RFC: https://github.com/graphql/graphql-spec/pull/825#issuecomment-1138335897.

Tests seem to fail because of some generated code. I (again?) have no idea how to solve that :sweat_smile:.


sunli829:
Thanks! 🙂

PR #1209, opened 2023-01-12
fix: Export directives to federation SDL so they can be composed.

dbanty:
Also added example of this feature in https://github.com/async-graphql/examples/pull/66

Also also updated the reported federation version to 2.1 so that `@defer` is available by default with Apollo tooling. The only change to 2.1 is `@composeDirective` which isn't applicable to this project since it isn't used with executable directives, only custom schema ones.

### DIFF
--- a/src/registry/export_sdl.rs
+++ b/src/registry/export_sdl.rs
@@ -113,9 +113,12 @@ impl Registry {
 
         if options.federation {
             writeln!(sdl, "extend schema @link(").ok();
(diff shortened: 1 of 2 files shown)


dbanty:
I think the format CI failure is unrelated to these changes

dbanty:
Ah, it was an earlier commit to examples that broke formatting. I made a PR for that too https://github.com/async-graphql/examples/pull/67

sunli829 (approved):
thanks! 🙂

Issue #1211, opened 2023-01-16
Dynamic Schema Does Not Resolve Correct Type for property `_services` (Federation Subgraphs)


nickisyourfan:
## Expected Behavior

Subgraph introspection of a federated server built with dynamic schema returns type `_Service`.

## Actual Behavior

Subgraph introspection of a federated server built with dynamic schema returns type `String`.

## Steps to Reproduce the Problem

1. Use dynamic schema.
2. Enable Federation.
3. Use Rover CLI to create supergraph -or- introspect the subgraph.
  - `rover supergraph compose http://localhost:5000`
  - `rover subgraph introspect http://localhost:5000`
4. Field `_service` returns `sdl` as type `String` causing subgraph introspection error. Note, regular introspection is valid using `rover graph introspect http://localhost:5000`

```
_service: "..."  
```

5. It should return type `_Service` instead, which is an object that contains the property `sdl` as a string.

```
_service: {
  sdl: "..."
}
```

## Specifications

- Version: `async-graphql = { version = "5.0.5", features = ["dynamic-schema"] }`
- Platform: Linux, Pop OS 22.04 LTS 
- Subsystem:

nickisyourfan:
This PR did provide the fix for me. @bbigras 

Would you like that I hit the close button or would you like an official contributer to close this?

PR #1212, opened 2023-01-16
fix: provide correct type for _service with dynamic schema

nickisyourfan:
Reported in issue #1211,

Performing subgraph introspection (federation) on a server built with dynamic schema results in incorrect type for `_service`. 

Steps to reproduce are within the issue ticket. 

Solution is modeled after code used to resolve the `_service` property when using non-dynamic schemas.


sunli829:
Thanks! 🙂

Issue #1001, opened 2022-07-30
Connections creates a `nodes` field as well as an `edges` and `pageInfo` field


SorenHolstHansen:
I am not really sure if this is intentional, but it seems weird.

When i use the built in cursor connection, i expect to have two top-level fields to query, namely `edges` and `pageInfo`, however, it seems that a new `nodes` field has been added that returns, well, the list of nodes. Is this intentional?

## Specifications

- Version: 4.0.4, using actix-web

sunli829:
This is intentional, see #952 🙂

xoac:
According to the documentation linked in #952 

> This leads to a potential usability improvement for connections; instead of just exposing the edges list, we **could** also expose a dedicated list of just the nodes, to avoid a layer of indirection.

So this means that `nodes` should be optional the simplest solution I came up would be to add feature "connection-without-nodes"  that will disable this endpoint from schema. @sunli829  would you accept nodes to be optional? Is my solution acceptable so I could prepare PR?

sunli829:
@xoac welcome to PR 🙂

PR #1215, opened 2023-01-19
feat: allow optionally enable nodes field on Connection strucutre with feature. Make this default because of sem ver

xoac:
I am not sure if this is correct solution in case of feature unification: https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

This is related to #1001

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,10 +17,17 @@ apollo_persisted_queries = ["lru", "sha2"]
 apollo_tracing = ["chrono"]
 email-validator = ["fast_chemail"]
 cbor = ["serde_cbor"]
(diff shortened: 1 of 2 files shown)


xoac:
This PR is breaking change I think as the existing code that depends on this with `default-feature = false` will have changed results. 

The solution for this could be `connection-type-remove-nodes-field` feature.

@sunli829  maybe this should be resolved in different way? For example similar to additional fields:
- https://docs.rs/async-graphql/5.0.5/src/async_graphql/types/connection/connection_type.rs.html#125-130

This will be simpler solution from user perspective and can keep Connection without any breaking changes.

sunli829:
Thanks! 🙂

PR #1218, opened 2023-01-19
feat: modify Connection to allow optionally disable nodes field in gql output. Keep nodes field enabled by default.

xoac:
This is different approach of #1215

In case of next major release would be good to consider returning `edges` field through `additional_fields` (If this will be possible because of reference to edges). This would simplify Connection structure

### DIFF
--- a/src/types/connection/connection_type.rs
+++ b/src/types/connection/connection_type.rs
@@ -9,6 +9,8 @@ use crate::{
     Object, ObjectType, OutputType, TypeName,
 };
 
+use super::{DisableNodesField, EnableNodesField, NodesFieldSwitcherSealed};
+
 /// Connection type
 ///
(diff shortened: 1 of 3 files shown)


sunli829:
I merged #1215 because it seemed simpler. 🙂

xoac:
Diff between #1218 and #1215:
  + 1218 is not breaking change (according to cargo feature documentation and sem-ver)
  + `Connection` with and without nodes can be used in the same code
  + if our crate will depends on async-graphql indirectly we can lose control over enabling/disabling nodes
  - 1218 complicates already complicated Connection strucutre
  
  I don't have strong opinion what is better. In my case both solution will works.

sunli829:
> Connection with and without nodes can be used in the same code

You're right, I didn't think about that. 🙂

PR #1237, opened 2023-02-15
disable default features for async-graphql in workspace dependencies

lazareviczoran:
All the other crates (in the `integrations` dir) that are referencing the parent crate (async-graphql) in the workspace have it turned off in their own Cargo.toml files, but since it's defined with the default features on the workspace level, and it's inherited in the other crates, specifying the `default-features = false` in the integration crates toml files doesn't have any effect.

Therefore, moving the `default-features = false` flag on the `async-graphql` dep in the `[workspace.dependencies]` instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,7 +135,7 @@ members = [
 ]
 
 [workspace.dependencies]
-async-graphql = { path = ".", version = "5.0.6" }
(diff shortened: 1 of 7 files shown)


lazareviczoran:
proposal to resolve #1239

sunli829:
thanks! 🙂

PR #1264, opened 2023-04-01
fix/delete-duplicate-conditional

nickisyourfan:
This conditional statement is declared on line 266 and 277. I removed the duplicate.


sunli829:
Thanks! 🙂

PR #1293, opened 2023-05-24
fix: prevent input check stack overflow

rtbunkerlw:
If a set of input objects has a local cycle that doesn't include the root type (i.e. the value of `current` in the code), the code can get stuck in recursion when trying to detect the cycle.

This change adds a hash table of visited types and avoids revisiting a type that's already been visited.


sunli829:
Thank you so much! 🙂

Issue #1281, opened 2023-04-27
README example is not compiling


alexanderankin:
## Expected Behavior
example from readme works:

<details><summary>example:</summary>

https://github.com/async-graphql/async-graphql/blob/62a8ee0643318287e87153d784e8bb958f5bcc94/README.md?plain=1#L29-L53

</details>

## Actual Behavior
compilation breaks with an error:

<details><summary>rustc error:</summary>

```
error[E0425]: cannot find value `graphiql` in this scope
  --> src/main.rs:24:38
   |
24 |   let app = Route::new().at("/", get(graphiql).post(GraphQL::new(schema)));
   |                                      ^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
error: could not compile `abc123` due to previous error
```
</details>

## Steps to Reproduce the Problem

1. `mkdir test && cd test && cargo init && vim src/main.rs`
2. `for pkg in async_graphql async_graphql_poem poem tokio ; do cargo add $pkg ; done`
3. also had to add the tokio annotation and enable the "full" feature in tokio

## Specifications

- Version: `5.0.7`
- Platform: `rustc 1.65.0 (897e37553 2022-11-02)`/`cargo 1.65.0 (4bc8f24d3 2022-10-20)`/`Linux $HOSTNAME 5.19.0-40-generic #41~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar 31 16:00:14 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
- Subsystem: not sure what is meant by subsystem

<details><summary>full `Cargo.toml`:</summary>

```toml
[package]
name = "abc123"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
async-graphql = "5.0.7"
async-graphql-poem = "5.0.7"
poem = "1.3.55"
tokio = { version = "1.28.0", features = ["full"] }
#async-graphql = "4.0.16"
```

</details>

full disclosure is that im new to rust and might be a user error, but might be of interest re: the first example of the library newcomers see

sburris0:
I have the same problem. The rustdoc book suggests including README within doctests which could prevent this:
https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#syntax-reference
```Rust
#[doc = include_str!("../README.md")]
#[cfg(doctest)]
[...]

PR #1301, opened 2023-06-04
Fix/improve README examples

sburris0:
Addresses #1281 

- Example for static schema compiles now
- Example for dynamic schema still doesn't compile, but is closer now
- Let `cargo fmt` do its thing
- Add note that dynamic schema requires `dynamic-schema` cargo feature

Fix details:
- Removing wildcard import disambiguates the use of `Result`
- Adding `graphiql` handler that uses `GraphiQLSource` fixes `graphiql` not being found (requires a few new imports)
- Adding `#[tokio::main]` prevents `main() can't be async` errors

I couldn't get [doc tests for README.md](https://github.com/async-graphql/async-graphql/issues/1281#issuecomment-1575299985) to work, but it may be worth trying again in the future

Thank you for the work you do :pray:

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,54 +28,72 @@ _This crate uses `#![forbid(unsafe_code)]` to ensure everything is implemented i
 ```rs
 use std::error::Error;
 
-use async_graphql::*;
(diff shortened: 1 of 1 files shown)


sunli829:
Thanks! 🙂

PR #1302, opened 2023-06-04
Add cursor getter to connection::Edge

meh:
I need it :panda_face:

### DIFF
--- a/src/types/connection/edge.rs
+++ b/src/types/connection/edge.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, marker::PhantomData};
+use std::{borrow::Cow, marker::PhantomData, ops::Deref};
 
 use crate::{
     connection::{DefaultEdgeName, EmptyFields},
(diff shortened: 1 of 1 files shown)


sunli829:
I've added `cursor` method to borrow the cursor value on the Edge, do you need `into_cursor` to get the ownership?

```rust
impl<Cursor, Node, Name> Edge<Cursor, Node, EmptyFields, Name>
where
    Cursor: CursorType + Send + Sync,
    Node: OutputType,
    Name: EdgeNameType,
{
    ....

    /// Returns the cursor of this edge
    pub fn cursor(&self) -> &Cursor {
        &self.cursor.0
    }
}
```

sunli829:
I've removed `CursorScalar` and exposed `Edge::cursor` 🙂

sunli829:
released in `v5.0.10`

meh:
Thanks a bunch!

PR #1319, opened 2023-07-20
Bump `indexmap` to `^2`

marcusirgens:
As far as I can tell, no code changes are required.

I added this to the release notes for `6.0.0`, but I don't think this requires a major versjon bump, so feel free to include it in a minor release.


sunli829:
Thanks! 🙂

PR #1324, opened 2023-07-22
chore: trim up some unnecessary code

Adam-Alj:
`struct_fields` is not used anywhere, this must be a relic from a previous implementation.

Cleaned up this small block to retain only the error functionality.


sunli829:
Thanks! 🙂

PR #1325, opened 2023-07-22
fix: fix incorrect sdl optionality with schema defaults

### LINKED ISSUE #1321: Incorrect Optionality in Generated SDL for InputObject Fields With Defined Defaults
## Expected Behavior

When generating the SDL for an InputObject with defined defaults, the defaultable field should not be marked as required (no '!' type).

The following InputObject: 
```
#[derive(InputObject)]
struct SampleInput {
    required_input: i32,

    #[graphql(default = 1)]
    optional_input: i32
}
```

Should generate the following SDL input definition:
```
input SampleInput {
	optionalInput: Int = 1
	requiredInput: Int!
}
```

## Actual Behavior

The following SDL is generated with the `optionalInput` field marked as required:
```
input SampleInput {
	optionalInput: Int! = 1
	requiredInput: Int!
}
```

Use the following code to reproduce:
```
use async_graphql::{Schema, EmptyMutation, EmptySubscription, SDLExportOptions};
use async_graphql_derive::{InputObject, Object};

struct Query;

#[Object]
impl Query {
    async fn sample(&self, _input: SampleInput) -> i32 {
        1 
    } 
}

#[derive(InputObject)]
struct SampleInput {
    required_input: i32,

    #[graphql(default = 1)]
    optional_input: i32
}

fn main() {
    let schema = Schema::build(Query, EmptyMutation, EmptySubscription).enable_federation().finish();

    let sdl_opts = SDLExportOptions::new().federation().sorted_arguments().sorted_fields();

    println!("{}", schema.sdl_with_options(sdl_opts));
}
```

## Specifications

- Version: 5.x, source
- Platform: MacOS/Ubuntu
- Subsystem: N/A

Adam-Alj:
Fixes: https://github.com/async-graphql/async-graphql/issues/1321

We need to know whether or not each type has a schema-default value in order to generate the SDL with correct optionality of fields. 

Say you have the following InputObject:
```
#[derive(InputObject)]
struct SampleInput {
    #[graphql(default = 1)]
    optional_input_defaulted: i32
}
```

The SDL currently generated by this is:
```
input SampleInput {
    optionalInputDefaulted: Int! = 1
}
```

Which means that callers are required to pass in a value `optionalInputDefaulted` as input, even though we are specifying that we are defaulting this value in the schema.

Then, in order to solve this issue, you would need to change your input object to the following:
```
#[derive(InputObject)]
struct SampleInput {
    #[graphql(default = 1)]
    optional_input_defaulted: Option<i32>
}
```

This generates the correct SDL of:
```
input SampleInput {
    optionalInputDefaulted: Int = 1
}
```
Which lets the caller make use of the defaulting behavior by not passing in a value.

----

By requiring types to be wrapped as Options even though they are being defaulted, we need to treat them as Options throughout the project, which is silly since we know that they are being defaulted and therefore are always `Some`. 

This change fixes that, by marking input fields with defaults as optional in the SDL.

### DIFF
--- a/derive/src/complex_object.rs
+++ b/derive/src/complex_object.rs
@@ -245,6 +245,7 @@ pub fn generate(
                     .map(|s| quote! {::std::option::Option::Some(::std::string::ToString::to_string(#s))})
                     .unwrap_or_else(|| quote! {::std::option::Option::None});
                 let default = generate_default(default, default_with)?;
+                let has_schema_default = default.is_some();
                 let schema_default = default
                     .as_ref()
                     .map(|value| {
@@ -265,7 +266,7 @@ pub fn generate(
                         args.insert(::std::borrow::ToOwned::to_owned(#name), #crate_name::registry::MetaInputValue {
                             name: ::std::string::ToString::to_string(#name),
                             description: #desc,
-                            ty: <#ty as #crate_name::InputType>::create_type_info(registry),
+                            ty: <#ty as #crate_name::InputType>::create_type_info(registry, #has_schema_default),
                             default_value: #schema_default,
                             visible: #visible,
                             inaccessible: #inaccessible,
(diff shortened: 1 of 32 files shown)


Adam-Alj:
Looks like CI checks are failing. I'll try and fix those up sometime tomorrow. The `tests/subscription_websocket_subscriptions_transport_ws.rs` test fails locally for me even off of master. Must have overlooked the failures due to it.

PR #1329, opened 2023-07-26
chore: make accessors public and reexport indexmap

broody:
Does it make sense to make the value accessors public? Currently, I'm duplicating this file in my own project to use them, but these are nice helpers I think the library could expose. Also, re-exporting the indexmap.

### DIFF
--- a/src/dynamic/value_accessor.rs
+++ b/src/dynamic/value_accessor.rs
@@ -157,6 +157,12 @@ impl<'a> ObjectAccessor<'a> {
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    /// Returns a reference to the underlying IndexMap
+    #[inline]
+    pub fn as_index_map(&'a self) -> &'a IndexMap<Name, Value> {
+        &self.0
+    }
 }
 
 /// A list accessor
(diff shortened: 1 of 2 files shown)


src/dynamic/value_accessor.rs
```diff
@@ -160,7 +160,7 @@ impl<'a> ObjectAccessor<'a> {
 }
 
 /// A list accessor
-pub struct ListAccessor<'a>(pub(crate) &'a [Value]);
+pub struct ListAccessor<'a>(pub &'a [Value]);
```
sunli829, line 163: add `ListAccessor::as_slice()` method?

src/dynamic/value_accessor.rs
```diff
@@ -106,7 +106,7 @@ impl<'a> ValueAccessor<'a> {
 }
 
 /// A object accessor
-pub struct ObjectAccessor<'a>(pub(crate) Cow<'a, IndexMap<Name, Value>>);
+pub struct ObjectAccessor<'a>(pub Cow<'a, IndexMap<Name, Value>>);
```
sunli829, line 109: add `ObjectAccessor::as_index_map()` method?

broody:
@sunli829 lemme know if there's anything else I should add

src/dynamic/value_accessor.rs
sunli829: keep `pub(crate)`

src/dynamic/value_accessor.rs
```diff
@@ -106,7 +106,7 @@ impl<'a> ValueAccessor<'a> {
 }
 
 /// A object accessor
-pub struct ObjectAccessor<'a>(pub(crate) Cow<'a, IndexMap<Name, Value>>);
```
sunli829, line 109: keep `pub(crate)`

PR #1331, opened 2023-07-31
Make TypeRefInner pub

banool:
Previously I made a PR where I made another type in this file pub. The reason was I have some code that builds GraphQL values dynamically rather than building a schema and querying it. I need this to be public for the same reason.

This is an example of the code that needs the type to be public: https://gist.github.com/banool/e4cf4397fa24dca6a9214a33ef991295.


banool:
This PR might provide more context on why I need this: https://github.com/aptos-labs/aptos-core/pull/9405.

Thanks!!

sunli829:
In version 6.0.1 I removed `TypeRefInner`.

banool:
Great, I just confirmed that 6.0.1 works for my purposes, thanks!

PR #1336, opened 2023-08-07
Parse "repeatable" in directive definitions

u9g:
Hi, thanks for all the great work that this library does! I just found myself needing to be able to parse the `repeatable` keyword, and it was easy enough to pr this upstream too!

Note that the `repeatable` keyword is in the spec [here](https://spec.graphql.org/October2021/#DirectiveDefinition)


sunli829:
Thanks! 🙂

PR #1342, opened 2023-08-11
Add english `#[graphql(inline)]` docs

### LINKED ISSUE #1335: Document `#[graphql(flatten)]` in the english version of the book
## Description of the feature

As per #1004 , async graphql supports the flatten attribute on Simple-/InputObject's, but this is not yet documented in the english version of the async-graphql book. I'd add it myself there, but I'm not sure which page to put it into, if someone tells me where to put it I'd be happy to add it myself. 

## Code example (if possible)

schitcrafter:
Fixes #1335

### DIFF
--- a/docs/en/src/define_input_object.md
+++ b/docs/en/src/define_input_object.md
@@ -109,3 +109,33 @@ pub struct CredentialsInput {
     password: String,
 }
 ```
+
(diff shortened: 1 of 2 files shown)


sunli829:
thanks! 🙂

PR #1346, opened 2023-08-14
Add support for `@interfaceObject` directive

### LINKED ISSUE #1238: Support for `@interfaceObject` and `@key` on interfaces
## Description of the feature

Add support for `@interfaceObject` and `@key` on interfaces that was added to the federation spec in v2.3. 

You can read more about this [here](https://github.com/apollographql/federation/issues/2277).

95ulisse:
Hello 👋 

This PR adds support for the [`@interfaceObject` directive](https://www.apollographql.com/docs/federation/federated-types/interfaces) on the `SimpleObject`, `Object` and `MergedObject` macros.

Example usage in Rust:

```rust
#[derive(SimpleObject)]
#[graphql(interface_object)] // <-- New attribute
struct MyInterfaceObject {
    pub id: u64,
}
```

GraphQL schema generated by the above example:

```graphql
#                                         v-- New directive
type MyInterfaceObject @key(fields: "id") @interfaceObject {
    id: Int!
}
```

The `@interfaceObject` directive is supported since Federation v2.3, so I had to also update the referenced Federation version used in the `@link` directives when exporting the schema, according to the [Apollo docs](https://www.apollographql.com/docs/federation/federated-types/interfaces#enabling-support). **I am not sure whether this is a breaking change or not!** I couldn't find any information in the official docs, but it probably is.

Closes #1238.


sunli829:
thanks! 🙂

PR #1353, opened 2023-08-23
Make `value` in `ValueAccessor` public

tusharmath:
This is usecase for us where we need to read the arguments as value without concerning about it's type. Eg

**Current Implementation**
```rust
fn get_arg<'a>(ctx: ResolverContext<'a>, name: String) -> &'a Value {
  let arg = ctx.args.get(name);

  // use a if-else ladder to check and convert from ValueAccessor to Value
  if let Ok(str) = arg..string() {
    Value::from(str).unwrap();
  } else if {
    // .. add more cases
  }
}
```

**Proposed Implementation**
```rust
fn get_arg<'a>(ctx: ResolverContext<'a>, name: String) -> &'a Value {
  ctx.args.get(name).0
}
```

This would be significantly faster for our usecase.

### DIFF
--- a/src/dynamic/value_accessor.rs
+++ b/src/dynamic/value_accessor.rs
@@ -6,7 +6,7 @@ use serde::de::DeserializeOwned;
 use crate::{Error, Name, Result, Value};
 
 /// A value accessor
-pub struct ValueAccessor<'a>(&'a Value);
+pub struct ValueAccessor<'a>(pub &'a Value);
 
 impl<'a> ValueAccessor<'a> {
     /// Returns `true` if the value is null, otherwise returns `false`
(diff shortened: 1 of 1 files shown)


sunli829:
Add a method called `into_inner`，what do you think?😄

tusharmath:
Sure! Happy to make that change.
What about `as_value` or something else with an `as` prefix?
I feel any `to` or `into` convey some sort of transformation, which generally is compute intensive. However, `as` is a conversion that in a way is coercing as type into something, without any overhead. Let me know what you think, can go with either.

tusharmath:
Done!

tusharmath:
When will it be released?

sunli829:
Waiting #1354

PR #1359, opened 2023-08-26
Fix panic in complexity visitor

grant0417:
I was getting a panic when using the complexity analysis and fragments at the same time, I found this line to be the one which panics. Not sure if this is the best solution but it prevents panics.


sunli829:
thanks! 🙂

PR #1367, opened 2023-09-08
Initialize 'source' Field in TracingExtension::parse_query to Allow Updates

kozaiy:
This pull request addresses an issue in the `TracingExtension::parse_query`. 
Previously, the 'source' field was not initialized, which prevented it from being updated later using the `.record()` method as per the [tracing documentation](https://docs.rs/tracing/0.1.38/tracing/span/struct.Span.html#method.record).

In this PR, I have added the initialization of the 'source' field using `tracinglib::field::Empty`, which allows it to be updated later.


sunli829:
thanks 🙂

PR #1376, opened 2023-09-18
added reset delay on load for dataloader

yasamoka:
This change allows the behavior of the dataloader delay to be changed. Instead of waiting for `delay` after the first new load (either at cold start or after all requested loads have been fulfilled), the timer is reset every time a load is requested. In my case, this has helped in a major way when a lot of loads are happening in succession (loading a huge number of objects). Previously, the batches would become very small with a short delay, while a longer delay risked waiting too long, especially when a smaller number of objects was requested instead.

The change is opt-in, requiring `reset_delay_on_load` to be set to `true`. Therefore, it shouldn't affect dataloader functionality for anyone who was using an older version of `async_graphql` or doesn't want this feature to be enabled.

All tests in `dataloader.rs` passed. I was unable to build the tests in the `tests` directory due to broken dependencies.

Feedback would be much appreciated as I'm still somewhat of a beginner in Rust, both language and ecosystem.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ apollo_tracing = ["chrono"]
 email-validator = ["fast_chemail"]
 cbor = ["serde_cbor"]
 chrono-duration = ["chrono", "iso8601"]
-dataloader = ["futures-timer", "futures-channel", "lru"]
+dataloader = ["futures-timer", "futures-channel", "lru", "tokio", "tokio-util"]
 decimal = ["rust_decimal"]
 default = ["email-validator", "tempfile", "playground", "graphiql"]
 password-strength-validator = ["zxcvbn"]
@@ -104,6 +104,7 @@ uuid08 = { version = "0.8", package = "uuid", optional = true, features = [
   "serde",
 ] }
 tempfile = { version = "3.2.0", optional = true }
+tokio-util = { version = "0.7.8", optional = true }
 
 # Non-feature optional dependencies
 blocking = { version = "1.0.2", optional = true }
--- a/src/dataloader/mod.rs
+++ b/src/dataloader/mod.rs
@@ -63,6 +63,7 @@ use std::{
     borrow::Cow,
     collections::{HashMap, HashSet},
     hash::Hash,
+    ops::DerefMut,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc, Mutex,
@@ -75,6 +76,8 @@ use fnv::FnvHashMap;
 use futures_channel::oneshot;
 use futures_timer::Delay;
 use futures_util::future::BoxFuture;
+use tokio::select;
+use tokio_util::sync::CancellationToken;
 #[cfg(feature = "tracing")]
 use tracing::{info_span, instrument, Instrument};
 #[cfg(feature = "tracing")]
@@ -128,6 +131,7 @@ pub trait Loader<K: Send + Sync + Hash + Eq + Clone + 'static>: Send + Sync + 's
 
 struct DataLoaderInner<T> {
     requests: Mutex<FnvHashMap<TypeId, Box<dyn Any + Sync + Send>>>,
+    token: Mutex<CancellationToken>,
     loader: T,
 }
 
@@ -185,6 +189,7 @@ pub struct DataLoader<T, C = NoCache> {
     inner: Arc<DataLoaderInner<T>>,
     cache_factory: C,
     delay: Duration,
+    reset_delay_on_load: bool,
     max_batch_size: usize,
     disable_cache: AtomicBool,
     spawner: Box<dyn Fn(BoxFuture<'static, ()>) + Send + Sync>,
@@ -199,10 +204,12 @@ impl<T> DataLoader<T, NoCache> {
         Self {
             inner: Arc::new(DataLoaderInner {
                 requests: Mutex::new(Default::default()),
+                token: Mutex::new(CancellationToken::new()),
                 loader,
             }),
             cache_factory: NoCache,
             delay: Duration::from_millis(1),
+            reset_delay_on_load: false,
             max_batch_size: 1000,
             disable_cache: false.into(),
             spawner: Box::new(move |fut| {
@@ -221,10 +228,12 @@ impl<T, C: CacheFactory> DataLoader<T, C> {
         Self {
             inner: Arc::new(DataLoaderInner {
                 requests: Mutex::new(Default::default()),
+                token: Mutex::new(CancellationToken::new()),
                 loader,
             }),
             cache_factory,
             delay: Duration::from_millis(1),
+            reset_delay_on_load: false,
             max_batch_size: 1000,
             disable_cache: false.into(),
             spawner: Box::new(move |fut| {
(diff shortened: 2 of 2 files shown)


yasamoka:
Is anyone going to take a look at this anytime soon?

sunli829:
I will refactor the loading logic of DataLoader, but thanks for the PR anyway. 🙂

yasamoka:
Is there a progress roadmap somewhere so that we avoid implementing what is already in the works? Thanks.

sunli829:
I'm thinking about how to remove the loading delay, but I haven't figured out how to do it yet. 🙂

yasamoka:
I have also been thinking about this over the past few days.

According to what I have gathered, there are 3 approaches that can be taken with dataloaders:

#### 1) Reference

This is the approach taken by Facebook's [dataloader](#https://github.com/graphql/dataloader) and the dataloader implemented in this library. A tick / delay timer fires off and the dataloader gathers whatever keys were loaded so far.

Assuming the dataloader is not tied to each request - which it is not in this framework - the benefit of this approach is with a high number of concurrent requests ensuring a constant flow of loaded keys within an order of magnitude of the intended max batch size. If there is a near-linear relationship between batch size and database latency, then a nice balance between request latency and database throughput is reached. Otherwise, one suffers at the expense of the other - either latency is allowed to increase in order to increase batch size and improve database throughput, or batch size is decreased in order to decrease request latency by resolving early at the expense of database throughput.

If the dataloader were to be tied to each request, and assuming database latency is the biggest factor contributing to request latency, then firing off every tick / delay breaks up the batches into small ones unnecessarily and leads to poor database throughput, causing higher request latency. This leads to the *raison d'être* of second approach.

#### 2) Reset delay on load

Instead of waiting for a constant tick / delay, we wait for an incremental delay. That way, as long as there are key loads, the timer resets and the dataloader does not break up the batches into small ones. We can keep the delay short enough to allow for all resolvers dependent on the dataloader to await on loading without spending too much time in idle, but long enough to ensure that a load can follow a previous one without the dataloader's delay firing off a premature batch load.

Assuming database latency is the biggest factor contributing to request latency, and optimizing for request latency, this approach was easy to implement in this PR and has completely solved the problem of breaking up batches into small ones for my use case.

However, the problem with this approach is that the lack of request awareness means that the delay can be reset by simultaneous / subsequent requests, causing all request latencies to increase until, in the worst case, max batch size is met and the dataloader finally fires off, satisfying resolvers that are now waiting across multiple requests.

Adding request awareness to this approach, in order to sidestep this problem, would mean that dataloaders now at most add `delay` to the (best case?) scenario of having all resolvers awaiting their dataloaders followed by each dataloader firing off exactly once per request / max batch size. However, if effort were spent on making dataloaders request-aware, I think it would rather be better to add that as part of the third approach.

#### 3) Register with dependent resolvers

This approach is intended to optimize for request latency, assuming database latency is the biggest factor contributing to that, by ensuring that dataloaders fire latest, as long as max batch size is not met, without having to wait for a delay that may be too small / too large.

In this scenario, he dataloader should fire off when all resolvers dependent on a dataloader are awaiting the dataloader. Additionally, the dataloader has to be made request-aware since we only have full information on in-flight requests - in other words, we do not know when a second request is coming in the future in order to decide whether we should include it or not, so we cannot wait for too little / too long. Either each request is taken on its own, or - if request latency optimization is not the primary target - only requests that are still in-flight can be taken into account and bundled together.

I am going to focus on the dataloader being tied to each singular request on its own.

I am not sure if Rust currently allows the semantics I am going to provide here since I have yet to dabble with macros and such, so bear with me if there are any inaccuracies.

```rust
struct SomeQueryData {
    ...
};

struct SomeDataLoader;

impl RegisteredLoader for SomeDataLoader {
    type Value = SomeQueryData;
    type Error = Error;

    fn load(&self, keys: &[i32]) -> Result<HashMap<i32, Self::Value>, Self::Error> {
        ...
    }
}

struct SomeData;

struct SomeObject;

#[Object]
impl SomeObject {
    pub async fn some_field(
        &self,
        // this registers the dataloader with the resolver
        // and increments the number of registered resolver
        // calls in this level of query execution
        dataloader: &SomeDataLoader,
    ) -> Result<SomeData> {
        // this increments the number of resolvers
        // awaiting the dataloader to fire off and
        // can only be called once
        let data = dataloader.load_one(self.id).await;
        ...
    }
}
```

For every level of query execution, which starts after all previously called resolvers have resolved and returned the data required, the number of resolvers to be executed is known ahead of time. This means that the number of resolvers registering a dataloader is also known ahead of time. This count is stored and compared against the number of resolvers that have called `load_one` or `load_many` on that dataloader. When the counts match, this means that all resolvers that have registered with the dataloader are now awaiting it to fire off, and the dataloader no longer needs to wait and can fire off immediately.

In case a resolver needs to call `load_one` or `load_many` for a dataloader multiple times, it can register the same dataloader multiple times instead, in order for count matching to still work, in one of two ways:

##### Data-dependence
```rust
pub async fn some_field(
        &self,
        dataloader1: &SomeDataLoader,
        dataloader2: &SomeDataLoader,
    ) -> Result<SomeData> {
        let data1 = dataloader.load_one(self.id).await;
        let data2 = dataloader.load_one(data1.id).await;
        ...
    }
```

##### Data-independence
```rust
pub async fn some_field(
        &self,
        dataloader1: &SomeDataLoader,
        dataloader2: &SomeDataLoader,
    ) -> Result<SomeData> {
        let future1 = dataloader.load_one(self.id);
        let future2 = dataloader.load_one(data1.id);
        let (data1, data2) = join!(future1, future2).await;
        ...
    }
```

Of course, this goes for registering multiple different dataloaders as well.

If a resolver uses a dataloader conditionally, then it should call, say, `ignore()` on the dataloader if it does not need to load in order for the counts to match. Otherwise, this would become a deadlock.

```rust
pub async fn some_field(
    &self,
    dataloader: &SomeDataLoader,
) -> Result<SomeData> {
    if (...) {
        dataloader.load_one(self.id).await;
        ...
    }
    else {
        dataloader.ignore();
    }
}
```

If this whole idea works, then given this dataloader has to be tied to each request, it should probably be a separate kind of dataloader - a `RegisteredDataLoader`, if you will.

Are there any gaps in my understanding that this should be possible? I, unfortunately, lack the knowledge to implement this at this time, although I realize that perhaps a few changes may have to happen first:

1) `RegisteredDataLoader`s should be registered to the schema in a way different to regular data (`.data()`), with a closure, in order to tie it to every request
2) `RegisteredDataLoader`s are now coupled to GraphQL schema execution so that counts can be communicated and reset
3) It should be ensured that resolvers that register dataloaders have to call either `load_one`, `load_many`, or `discard` exactly once. Maybe there is a way to enforce this with Rust's type system and move semantics?

Let me know what you think.

sunli829:
Thank you very much for your suggestion, I can see that you have a very deep understanding of GraphQL! What I want to achieve right now is the third one, we need to allow users to send an immediate query notification to `DataLoader` at the end of each level of query, or as you said add a `RegisteredDataLoaders`, which is automatically send query notification by the framework.

It’s not hard to do these things, we need to find a more concise way of using it, and try to avoid breaking changes. 🙂

yasamoka:
Glad to read that it's not hard!

I can try and help if I get a solid grasp of the macro system and async, if you would lay out the needed changes / experiments / features etc...

PR #1378, opened 2023-09-19
Add support for entities without a reference resolver

akrantz01:
Resolves #1361 

Heya :wave:

This PR adds support for defining entities without a reference resolver using the [`resolvable` parameter of the `@key` directive](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#key) on the `SimpleObject` and `Object` macros.

The key to be marked unresolvable can be either implicitly (`#[graphql(unresolvable)]`) or explicitly (`#[graphql(unresolvable = "<key>")]`) defined. When implicitly defined, the macro assumes all non-skipped fields are part of the key. When explicitly defined, the macro assumes the provided key is valid for the object. The explicit key will most likely be used for entities referenced by compound keys with nested fields, which cannot be easily validated.

This should not be a breaking change. The `resolvable` argument defaults to `true`, so it will not affect schema generation unless explicitly requested.

### Example usage:

```rust
#[derive(SimpleObject)]
#[graphql(unresolvable)] // <-- new attribute
struct ImplicitObject {
	pub id: u64,
	pub username: String,
}

#[derive(SimpleObject)]
#[graphql(unresolvable = "id")] // <-- new attribute
struct ExplicitObject {
    pub id: u64,
}
```
 
This example will generate the following schema:

```graphql
#                                               v-- added directive argument
type ImplicitObject @key(fields: "id username", resolvable: false) {
    id: Int!
    username: String!
}

#                                      v-- added directive argument
type ExplicitObject @key(fields: "id", resolvable: false) {
    id: Int!
}
```

These objects will also not be added to the `_Entity` union as they cannot be resolved by the subgraph.


sunli829:
thanks! 🙂

Issue #1279, opened 2023-04-19
Passing data to resolvers per item in subscription stream


bakhtiyarneyman:
## Description of the feature
When doing queries/mutations there is a straightforward way to pass per-request data (`SchemaBuilder::build(...).data(my_data)`). For subscriptions there is a way to pass data _per request_, however there is no way to pass data _per stream item_ returned from the server. 

An example of why this would be useful is a `Transaction` object, that can be used to make sure that each message returned via subscription show a consistent view of the data.

## Possible API

```
#[Subscription]
impl Subscription {
    async fn comment<F>(&self, ctx: &Context<'_>, comment_id: i32) -> impl Stream<Item = F> 
        where F: FnOnce() -> Comment {
        context
            .data_unchecked::<CommentStream>()
            .map(|comment| {
                context.insert_subscription_data(ctx.data_unchecked::<Pool>().begin());
                comment
            }) 
    }
}
```

bakhtiyarneyman:
It probably would be simpler to change `with_data` of `GraphQLWebSocket` so that instead of accepting `Data` it accepts a callback that returns a `BoxFuture` that returns `Data`.

bakhtiyarneyman:
Specifically for transactions there should also be a way to run a callback after all the resolvers have run (to commit it). Using Drop on a guard type is non-ideal because the commit may fail.

sunli829:
This is what you want, right?

https://github.com/async-graphql/async-graphql/blob/6684cd80b92275c278550fdde8e1e11161686139/tests/subscription.rs#L98

bakhtiyarneyman:
@sunli829 Hey, really curious to learn your thoughts on supporting this functionality.

PR #1399, opened 2023-10-14
Per-message hooks: one runs before the resolvers, one after. 

bakhtiyarneyman:
This is useful for making sure that all resolvers for a message emitted in subscription context run in the same transaction.
Closes [this issue](https://github.com/async-graphql/async-graphql/issues/1279).

### DIFF
--- a/derive/src/subscription.rs
+++ b/derive/src/subscription.rs
@@ -311,6 +311,29 @@ pub fn generate(
                         let field = ::std::clone::Clone::clone(&field);
                         let field_name = ::std::clone::Clone::clone(&field_name);
                         async move {
+                            let message_data = if let ::std::option::Option::Some(f) = query_env.per_message_pre_hook.as_ref() {
+                                match (*f)().await {
+                                    ::std::result::Result::Ok(md) => {
+                                        ::std::option::Option::unwrap_or_else(
+                                            md, || ::std::default::Default::default()
+                                        )
+                                    },
+                                    ::std::result::Result::Err(e) => {
+                                        return ::std::result::Result::Err(
+                                            e.into_server_error(pos)
+                                            .with_path(::std::vec![
+                                                #crate_name::PathSegment::Field(
+                                                    ::std::borrow::ToOwned::to_owned(
+                                                        &*field_name
+                                                    )
+                                                )
+                                            ])
+                                        )
+                                    },
+                                }
+                            } else {
+                                ::std::default::Default::default()
+                            };
                             let ctx_selection_set = query_env.create_context(
                                 &schema_env,
                                 ::std::option::Option::Some(#crate_name::QueryPathNode {
@@ -318,6 +341,7 @@ pub fn generate(
                                     segment: #crate_name::QueryPathSegment::Name(&field_name),
                                 }),
                                 &field.node.selection_set,
+                                &message_data,
                             );
 
                             let execute_fut = async {
@@ -347,6 +371,24 @@ pub fn generate(
 
                                 use ::std::iter::Extend;
                                 resp.errors.extend(::std::mem::take(&mut *query_env.errors.lock().unwrap()));
+                                if let ::std::option::Option::Some(f) = query_env.per_message_post_hook.as_ref() {
+                                    match (*f)(&message_data).await {
+                                        ::std::result::Result::Ok(md) => {
+                                        },
+                                        ::std::result::Result::Err(e) => {
+                                            resp.errors.push(
+                                                e.into_server_error(pos)
+                                                .with_path(::std::vec![
+                                                    #crate_name::PathSegment::Field(
+                                                        ::std::borrow::ToOwned::to_owned(
+                                                            &*field_name
+                                                        )
+                                                    )
+                                                ])
+                                            )
+                                        },
+                                    }
+                                }
                                 resp
                             };
                             #crate_name::futures_util::pin_mut!(execute_fut);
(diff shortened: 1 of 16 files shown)


bakhtiyarneyman:
@sunli829 Friendly ping.

bakhtiyarneyman:
@sunli829 Gentle ping.

bakhtiyarneyman:
@sunli829 Gentle ping.

bakhtiyarneyman:
@sunli829 Gentle ping. It's been almost 2.5+ months since this pull request has been opened. It would be great to see it going forward sooner than later. In particular, we would like to open-source our internal crate which doesn't make much sense to publish without this change. Can we at least get some response regarding why this PR hasn't been merged and how we can help accelerate the process, please?  

Thank you!

bakhtiyarneyman:
@sunli829 Gentle ping.

bakhtiyarneyman:
@sunli829 Gentle ping.

bakhtiyarneyman:
@sunli829 Gentle ping!

bakhtiyarneyman:
@sunli829 Gentle ping!

xamgore:
Hi! Why not just use a patched version then?

```toml
[patch.crates-io] # cargo update
async-graphql = { git = "https://github.com/async-graphql/async-graphql.git", rev = "refs/pull/1399/head" }
```

bakhtiyarneyman:
Hi! Thank you for your suggestion!

#### What we want to publish 

Internally we use our own fork of `async_graphql`. The library that we have built on top of it is compiling GraphQL queries to singular Postgres SQL statements Hasura-style. This means that for deeply nested queries there is no need to make multiple queries for each level of nesting! It's just one query, with small latency. 

#### Why do we need per-message hooks 
Our library strives to be minimal, so we punt on authorization. The recommended way to handle it is to use RLS (which admittedly is not everybody's cup of tea). But to use RLS one needs to make sure that each of the SQL queries of a GraphQL transaction run in a transaction that sets appropriate variables early on. 

#### Why do we need to update `async-graphql`
Now, not everything resides in a database, so in addition to using our crate the user should be able to use `async_graphql` directly as well. This means that the user will need to import both at the same time. Moreover, because this feature would be useful to anyone relying on RLS, we believe that it's perfectly in-scope for async-graphql.

 
#### Why we think `patch` approach is non-ideal

I have not used `patch` feature extensively, but my understanding is:
- The patch needs to be introduced both at the definition of our package, as well as the user's local crate definition. That's leads to additional friction for the user, as well as cryptic errors when there is a version skew in the patch revision.   
- Issues like [this](https://github.com/rust-lang/cargo/issues/10382) suggest that patch behavior is not obvious to users.
-  Not all tooling supports these advanced cargo features (I'm looking at you, `nix`). 

#### Lack of communication
We could potentially take the patch/fork option, but we are still hopeful that this pull request gets merged. We don't see a meaningful cost to merging this, and it benefits not just us, but everybody who cares about data integrity or RLS. But to know how to continue we would need to understand, why isn't it merged yet. 
- Is the motivation for this feature not articulated sufficiently?
- Is the current implementation somehow flawed?
- Is the implementation incompatible with the future development that the library maintainers envision?
- Is it just a bandwidth issue on the PR review? Everybody is very busy, of course so it's understandable.
etc.   

Unfortunately, communication on this issue has been non-existent, until your comment Igor, so thank you for that. A little bit of clarity from the maintainers goes a long way!

sunli829:
Providing two callback functions is weird and I don't think it's convenient. I need to think about whether there is a better solution.🙂

sunli829:
Why don't you use `Extension`? Only need to implement `execute_start` and `execute_end` to achieve your goal.

sunli829:
```rust
struct MyExtension {
}

#[async_trait::async_trait]
impl Extension for MyExtensionImpl {
    async fn execute(
        &self,
        ctx: &ExtensionContext<'_>,
        operation_name: Option<&str>,
        next: NextExecute<'_>,
    ) -> Response {
        // do something before
        ...

        let res = next.run(ctx, operation_name).await;

        // do something after
        ...
        
        res
    }
}

let schema = Schema::build(Query, EmptyMutation, Subscription)
    .extension(MyExtension {
    })
    .finish();
```

bakhtiyarneyman:
@sunli829 Thank you for responding! An `Arc<Transaction>` needs to be passed as data, so that resolvers can use it. Seemingly, `ExtensionContext` can only be used to _get_ data. So we can't use extensions for this purpose. Please correct me if this is not correct.

sunli829:
I just need to add a way for Extension::execute to set the context data, and the problem is solved, right?

bakhtiyarneyman:
I _think_ that should work... Here are the benefits of your approach.
- There's going to be less code emitted by the macros.
- Less code to write when writing integrations with warp, axum, and such.
- No need for the user to write similar, but slightly different code for requests and subscription messages.

The costs are harder for me to assess (e.g. will the latency increase linearly in number of extensions for each access?)

sunli829:
I added `NextExecute::run_with_data` to attach context data before execution, here is the example (from this PR), can you help take a look? 🙂

https://github.com/async-graphql/async-graphql/blob/42cb57f4c58c85679f9926d05ef4cab905e1d82c/tests/extension.rs#L388

bakhtiyarneyman:
Thanks, Sunli! I started looking into it but hit an unrelated issue, will report in a few days.

bakhtiyarneyman:
Can confirm it works!

sunli829:
`v7.0.3` has been released, this PR can be closed. 🙂

joleeee:
@bakhtiyarneyman did you also manage to integrate postgres transactions into Dataloader, or did you not use Loader<>? To use RLS

PR #1405, opened 2023-10-24
fix check for serial

sujeetsr:
Fix the serial check for query execution in dynamic schema


tusharmath:
@sunli829 This is a major logical bug in AsyncGraphQL, can we make a release with this fix?

sunli829:
thanks! 🙂

tusharmath:
Thanks! When can we have a release for this?

PR #1420, opened 2023-11-13
Run batch requests concurrently

sujeetsr:
Enable batch requests in dynamic schema to run concurrently. This is just adding the same fix in [411239f4b35e1e72e8cedb6fcc834cc05b4e301a](https://github.com/async-graphql/async-graphql/commit/411239f4b35e1e72e8cedb6fcc834cc05b4e301a) to executor so that it applies to dynamic schema too.

@sunli829 , is the `execute_batch` function in `src/schema.rs` not needed anymore with this change, because that implements `Executor` too?

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use futures_util::{stream::BoxStream, StreamExt};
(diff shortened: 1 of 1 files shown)


sunli829:
Thanks

PR #1428, opened 2023-11-23
Feature extend ResolveInfo with field attribute

vstepanyuk:
Extend ResolveInfo with field attribute, to have more context information in extensions


src/extensions/mod.rs
```diff
@@ -128,6 +129,9 @@ pub struct ResolveInfo<'a> {
 
     /// If `true` means the current field is for introspection.
     pub is_for_introspection: bool,
+
+    #[doc(hidden)]
```
sunli829, line 133: remove `#[doc(hidden)]` ?

Issue #1430, opened 2023-11-27
Support axum 7


ziimakc:
Axum was updated to [version 7](https://github.com/tokio-rs/axum/blob/main/axum/CHANGELOG.md) which doesn't work well with async-graphql.

## Expected Behavior
No errors.

## Actual Behavior
```
let app = Router::new().route( "/", get(playground).post_service(GraphQL::new(gql_schema.schema)));
/*
the trait bound `GraphQL<async_graphql::Schema<graphql::schema::Query, graphql::schema::Mutation, EmptySubscription>>: tower_service::Service<axum::http::Request<Body>>` is not satisfied
the trait `tower_service::Service<http::request::Request<B>>` is implemented for `GraphQL<E>` */
```

## Steps to Reproduce the Problem

1. Clone [example](https://github.com/async-graphql/examples/blob/966d899c65a1255be4a6f85d0a825a945e4bbb93/axum/starwars/src/main.rs).
2. Update to axum 7
3. Run cargo check

## Specifications

- Version: axum 0.7.1
- Platform: windows
- Subsystem: wsl 2/ubuntu 22

azzamsa:
Then we need to update the example https://github.com/async-graphql/examples/tree/master/axum

PR #1431, opened 2023-11-27
Update axum and http

davidpdrsn:
Updates to axum 0.7 and http 1.0.

Note that this is a breaking change because http is a public dependency of async-graphql.

Some integrations, such as warp, can't be updated yet because they don't support http 1.0. Might also get issues with actix-web which appears to use http 0.2 🤔 

Fixes https://github.com/async-graphql/async-graphql/issues/1430

### DIFF
--- a/integrations/axum/src/extract.rs
+++ b/integrations/axum/src/extract.rs
@@ -2,13 +2,10 @@ use std::{io::ErrorKind, marker::PhantomData};
 
 use async_graphql::{futures_util::TryStreamExt, http::MultipartOptions, ParseRequestError};
 use axum::{
-    body::HttpBody,
-    extract::{BodyStream, FromRequest},
-    http::{self, Method, Request},
+    extract::{FromRequest, Request},
+    http::{self, Method},
     response::IntoResponse,
-    BoxError,
 };
-use bytes::Bytes;
 use tokio_util::compat::TokioAsyncReadCompatExt;
 
 /// Extractor for GraphQL request.
@@ -29,25 +26,25 @@ impl<R> GraphQLRequest<R> {
 pub mod rejection {
     use async_graphql::ParseRequestError;
     use axum::{
-        body::{boxed, Body, BoxBody},
+        body::Body,
         http,
         http::StatusCode,
-        response::IntoResponse,
+        response::{IntoResponse, Response},
     };
 
     /// Rejection used for [`GraphQLRequest`](GraphQLRequest).
     pub struct GraphQLRejection(pub ParseRequestError);
 
     impl IntoResponse for GraphQLRejection {
-        fn into_response(self) -> http::Response<BoxBody> {
+        fn into_response(self) -> Response {
             match self.0 {
                 ParseRequestError::PayloadTooLarge => http::Response::builder()
                     .status(StatusCode::PAYLOAD_TOO_LARGE)
-                    .body(boxed(Body::empty()))
+                    .body(Body::empty())
                     .unwrap(),
                 bad_request => http::Response::builder()
                     .status(StatusCode::BAD_REQUEST)
-                    .body(boxed(Body::from(format!("{:?}", bad_request))))
+                    .body(Body::from(format!("{:?}", bad_request)))
                     .unwrap(),
             }
         }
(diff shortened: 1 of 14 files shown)


ruizdiazever:
@davidpdrsn great contribution, thanks!

integrations/axum/src/extract.rs
```diff
@@ -97,17 +91,14 @@ impl<R> GraphQLBatchRequest<R> {
...
-    B: HttpBody + Send + Sync + 'static,
-    B::Data: Into<Bytes>,
-    B::Error: Into<BoxError>,
     S: Send + Sync,
     R: IntoResponse + From<ParseRequestError>,
 {
     type Rejection = R;
 
-    async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
+    async fn from_request(req: Request, _state: &S) -> Result<Self, Self::Rejection> {
```
IvanUkhov, line 101: Not caught by the existing tests, but it fails for me:

```
102 |         if let (&Method::GET, uri) = (req.method(), req.uri()) {
    |                                       --- has type `&axum::http::Request<Body>` which is not `Send`
...
128 |                 .await?,
    |                  ^^^^^ await occurs here, with `req` maybe used later
...
132 |     }
    |     - `req` is later dropped here
help: consider moving this into a `let` binding to create a shorter lived borrow
```

A solution is to check the method without binding:

```rust
if req.method() == Method::GET {
    let uri = req.uri();
    …
```

IvanUkhov:
Continuing with my comment above, if you initialize the examples submodule and go to `examples/axum/starwars` and try to compile, you will get something like this:

```
   Compiling async-graphql-axum v6.0.11 (/Users/ivanukhov/Downloads/async-graphql/integrations/axum)
error: future cannot be sent between threads safely
   --> /Users/ivanukhov/Downloads/async-graphql/integrations/axum/src/extract.rs:101:86
    |
101 |       async fn from_request(req: Request, _state: &S) -> Result<Self, Self::Rejection> {
    |  ______________________________________________________________________________________^
102 | |         if let (&Method::GET, uri) = (req.method(), req.uri()) {
103 | |             let res = async_graphql::http::parse_query_string(uri.query().unwrap_or_default())
104 | |                 .map_err(|err| {
...   |
131 | |         }
132 | |     }
    | |_____^ future created by async block is not `Send`
    |
    = help: the trait `Sync` is not implemented for `(dyn HttpBody<Error = axum::Error, Data = bytes::Bytes> + std::marker::Send + 'static)`
note: future is not `Send` as this value is used across an await
   --> /Users/ivanukhov/Downloads/async-graphql/integrations/axum/src/extract.rs:128:18
    |
102 |         if let (&Method::GET, uri) = (req.method(), req.uri()) {
    |                                       --- has type `&axum::http::Request<Body>` which is not `Send`
...
128 |                 .await?,
    |                  ^^^^^ await occurs here, with `req` maybe used later
...
132 |     }
    |     - `req` is later dropped here
help: consider moving this into a `let` binding to create a shorter lived borrow
   --> /Users/ivanukhov/Downloads/async-graphql/integrations/axum/src/extract.rs:102:39
    |
102 |         if let (&Method::GET, uri) = (req.method(), req.uri()) {
    |                                       ^^^^^^^^^^^^
    = note: required for the cast from `Pin<Box<[async block@/Users/ivanukhov/Downloads/async-graphql/integrations/axum/src/extract.rs:101:86: 132:6]>>` to `Pin<Box<(dyn futures_util::Future<Output = Result<GraphQLBatchRequest<R>, R>> + std::marker::Send + 'async_trait)>>`

error: could not compile `async-graphql-axum` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
```

davidpdrsn:
Thanks for noticing! Wasn't aware of the submodule.

But before I fix that I'd like to know if updating now is even feasible given that it requires bumping http to 1.0 which is a public dependency of async-graphql. If that breaks the actix-web integration then that's quite unfortunate.

bkonkle:
I'm seeing the same thing as @IvanUkhov. His suggested fix in the `extract.rs` solves it for me.

I do see the `actix-web`, `poem`, and `warp` integrations breaking because of the http 1.0 change, though.

Actix is targeting their http upgrade for [v5.0](https://github.com/actix/actix-web/milestone/12) with [this ticket](https://github.com/actix/actix-web/pull/3208).

I couldn't find any Poem or Warp activity around updating `http` yet.

It works for Axum though, so in the meantime I'll point to my Github forks. 😁

sunli829:
@davidpdrsn use `cargo +nightly fmt` 🙂

sunli829:
waiting on `multer` release for `http 1.0`.

toxeus:
@sunli829 `multer` for `http 1.0` has been released: https://crates.io/crates/multer/3.0.0

IgnisDa:
@sunli829 When do you plan to make a new release?

sunli829:
released in `v7.0.0`

PR #1484, opened 2024-03-09
Change the type_name of EmptySubscription, fix #1435

### LINKED ISSUE #1435: Wrong type_name in EmptySubscription
https://github.com/async-graphql/async-graphql/blob/9086ca7ef0bf89723bbfe6d054e7f086b0b1cbf0/src/types/empty_subscription.rs#L16

Got a panic because I used the EmptyMutation and EmptySubscription at the same time, is it intentional that EmptySubscription has "EmptyMutation" as type_name?

xamgore:
See the linked issue.


sunli829:
thanks 🙂

PR #1489, opened 2024-03-13
fix: add `description` to `__schema` introspection result

YassinEldeeb:
Newer Graphiql versions' introspection queries include `description` to the `__schema` as the following:

```graphql
query IntrospectionQuery {
  __schema {
    description <------ required
    queryType {
      name
    }
    mutationType {
      name
    }
    subscriptionType {
      name
    }
    types {
      ...FullType
    }
    directives {
      ...
    }
  }
}
...
```

Introspection queries through newer versions of graphiql result in:
```sh
“Unknown field \“description\” on type \“__Schema\“. Did you mean \“subscriptionType\“?”
```

Official implementation reference:
https://github.com/graphql/graphql-js/blob/main/src/type/introspection.ts

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { rust: nightly-2022-11-02, os: ubuntu-20.04 }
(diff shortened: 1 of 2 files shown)


dotansimha (approved):

sunli829:
thanks 🙂

PR #1491, opened 2024-03-13
Fix validation of non-nullable variables with default values

### LINKED ISSUE #1486: Non nullable variables should allow default values
## Expected Behavior

Queries should be able to define non-nullable variables with default variables.  This is currently prevented by the check here:
https://github.com/async-graphql/async-graphql/blob/759fb0e594606aecf4a239b00d6f3f8aeb22d885/src/validation/rules/default_values_of_correct_type.rs#L32-L36

The formal specification for how variables are processed is described here: https://spec.graphql.org/October2021/#sec-Coercing-Variable-Values

In short, in GraphQL `non-nullable` means the value can't be null, but isn't directly related to weather or not the variable must be provided by the user:

This leads to some non-intuitive behavior:
* A non-nullable variable with a default value is optional, and will error if the variable is explicitly passed as null
* A nullable varaible with a non-null default value can still be explicitly set to null, overwriting the default value

This same logic applies to defaults provided in the schema for arguments (https://spec.graphql.org/October2021/#sec-Coercing-Field-Arguments)

The relevant part of the JS reference implementation is here: https://github.com/graphql/graphql-js/blob/main/src/execution/values.ts#L94-L105

This checks for a default value BEFORE before checking for non-nullablity, and will only error if the field is both non-nullable and does not provide a default value

## Actual Behavior

Any query that defines a default value for a non-nullable variable errors

1. Define a schema that has a non-nullable argument
2. Write a query with a non-nullable variable with a default value
3. Execute the query, without providing any variables

## Specifications

- Version:
- Platform:
- Subsystem:

eth3lbert:
Resolve #1486

### DIFF
--- a/src/validation/rules/default_values_of_correct_type.rs
+++ b/src/validation/rules/default_values_of_correct_type.rs
@@ -29,12 +29,7 @@ impl<'a> Visitor<'a> for DefaultValuesOfCorrectType {
         }
 
         if let Some(value) = &variable_definition.node.default_value {
-            if !variable_definition.node.var_type.node.nullable {
-                ctx.report_error(vec![variable_definition.pos],format!(
-                    "Argument \"{}\" has type \"{}\" and is not nullable, so it can't have a default value",
-                    variable_definition.node.name, variable_definition.node.var_type,
-                ));
-            } else if let Some(reason) = is_valid_input_value(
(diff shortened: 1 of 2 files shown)


hayes (approved):

sunli829:
thanks 🙂

PR #1492, opened 2024-03-14
feat: add registry method in `dynamic::Registry`

ssddOnTop:
A registry is used to implement check on functionality but it is private currently. Adding a `registry` method in dynamic::Schema will return a reference to registry instance which can be used for the same


sunli829:
thanks 🙂

PR #1493, opened 2024-03-15
Allow non-scalars to be used as directive arguments

kschat:
`TypeDirective` argument types seem to be erroneously limited to just scalars. Per the GraphQL spec, [`DirectiveDefinition`](http://spec.graphql.org/June2018/#sec-Type-System.Directives) defines [`ArgumentsDefinition`](http://spec.graphql.org/October2021/#ArgumentsDefinition) as `( InputValueDefinition )`, which allows argument types to be [`NamedType`, `ListType`, or `NonNullType`](http://spec.graphql.org/October2021/#Type).

I _believe_ `InputType` is the appropriate trait to use here, but if it's not please let me know and I can update this PR.

EDIT: I updated this PR to also support default arguments in `TypeDirective`s using the `#[graphql(default)]` attribute.

### DIFF
--- a/derive/src/type_directive.rs
+++ b/derive/src/type_directive.rs
@@ -108,7 +108,7 @@ pub fn generate(
 
         directive_input_args.push(quote! {
             if let Some(val) = #crate_name::InputType::as_raw_value(&#arg_ident) {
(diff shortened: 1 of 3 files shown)


sunli829:
thanks 🙂

PR #1499, opened 2024-03-19
handle null interface items

xuorig:
It's not currently possible to return `null` for a field returning an interface type in dynamic schemas. It fails since resolution expects a `WithType`.

This PR fixes this, but I haven't added tests / looks at whether this handles all cases correctly. Wanted your thoughts first. If that seems like the right fix, I'm happy to clean this PR up / add tests.

### DIFF
--- a/src/dynamic/resolve.rs
+++ b/src/dynamic/resolve.rs
@@ -601,6 +601,7 @@ async fn resolve_value(
             )
             .await
         }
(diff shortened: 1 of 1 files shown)


sunli829:
Only optional types can be `nil`, but `interface` cannot, so I think this change is incorrect and will result in a response that does not conform to the GraphQL specification.

PR #1500, opened 2024-03-19
Support Directives on Subscriptions

jeffutter:
This supports the same directives that are supported on Object on Subscriptions.

My main use-case here is to support WunderGraph EDFS, which requires directives on Subscription fields: https://cosmo-docs.wundergraph.com/router/event-driven-federated-subscriptions-edfs


xamgore:
Do you have any idea, why the pipeline checks could fail?

jeffutter:
@xamgore I believe I fixed this up by updating `parser/src/parse/generated.rs`

tests/type_directive.rs
```diff
@@ -50,10 +51,36 @@ pub fn test_type_directive_1() {
...
+
+    let schema = Schema::build(Query, EmptyMutation, Subscription)
+        .enable_subscription_in_federation()
+        .finish();
 
     let sdl = schema.sdl_with_options(SDLExportOptions::new().federation().compose_directive());
 
+    println!("----");
+    println!("{}", sdl);
+
```
xamgore, line 83: For me everything else looks fine except these probably forgotten debug statements.

jeffutter: Fixed. Thanks for catching it.

xamgore:
@sunli829 would you be so kind to pay your attention here, please?

sunli829:
sorry for the late processing, it's been a busy time. 😂

PR #1503, opened 2024-03-25
Updating dependencies to the actual version

negezor:
I'm constantly keeping track of the growing number of packages in my project, and I noticed an increase in duplicate versions of dependencies that `async-graphql` uses. So I took a little liberty and updated dependencies that have had a new version for a long time. I looked at the changelog of all the updated packages, it shouldn't break anything 😁


negezor:
This PR also requires updating opentelemetry in the examples. So I also made a PR to update the dependencies in the examples. https://github.com/async-graphql/examples/pull/82

sunli829:
thanks 🙂

PR #1504, opened 2024-03-25
Update lru dependency

negezor:
Update `lru` dependency that requires passing `NonZeroUsize` instead of just `usize`.


sunli829:
thanks 🙂

PR #1517, opened 2024-04-30
Added support for deploying to wasm targets with axum - (without subscriptions)

spigaz:
Added support for wasm targets, although at the expense of subscriptions support.

The reason is that there is a nested dependency in the axum implementation of websockets that requires mio and mio for now doesn't compile for wasm.

Nevertheless, its functional without subscriptions and allows developing a platform specific version of the subscriptions.


sunli829:
thanks! 🙂

PR #1524, opened 2024-05-18
Improve alignment of directive behavior with GraphQL spec

### LINKED ISSUE #1187: Oneof directive not included in introspection result
## Expected Behavior
I would expect the `@oneof` directive to be included in the result of an introspection query whenever the `#[derive(OneofObject)]` macro is used on a struct.

## Actual Behavior
The directive is not listed in the introspection query result:

image

1. Clone this minimal example: https://github.com/bvanneerven/introspection-example
2. Run with `cargo run`
3. Go to http://localhost:8001 and open the network tab in the browser developer tools
4. Trigger an introspection query by clicking the 'Re-fetch GraphQL schema' button
5. Verify that the `@oneof` directive is not listed in the response from the server

## Specifications

- Version: `{ git = "https://github.com/async-graphql/async-graphql" }`
- Platform: WSL (Ubuntu-20.04)

bramvanneerven:
Before, only the `skip` and `include` directives were included in schema introspection. This PR adds all built-in directives to the introspection result (always), and to the SDL export (only when they are used) to better align with the latest [GraphQL working draft specification](https://spec.graphql.org/draft/#sel-EAHnBPDLAADNC1Br_F).

Fixes #1187.

### DIFF
--- a/src/registry/export_sdl.rs
+++ b/src/registry/export_sdl.rs
@@ -139,8 +139,6 @@ impl Registry {
             writeln!(sdl, "{}", directive.sdl()).ok();
         });
 
-        writeln!(sdl).ok();
-
         if options.federation {
             writeln!(sdl, "extend schema @link(").ok();
             writeln!(sdl, "\turl: \"https://specs.apollo.dev/federation/v2.3\",").ok();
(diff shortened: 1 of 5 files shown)


bramvanneerven:
All tests pass now, and I added some tests to verify the correctness of the changes. 
Ready to merge if you approve of the changes @sunli829.

sunli829:
thanks! 🙂

Issue #1526, opened 2024-05-25
Default Argument Values Not Applied When Arguments Are Omitted


be-hunter:
When creating a GraphQL schema using async_graphql, default argument values are not being applied when arguments are omitted from the query. The following example demonstrates the issue:

```graphql
use async_graphql::{dynamic::*, value, Value};

let query = Object::new("Query").field(
    Field::new("test", TypeRef::named_nn(TypeRef::INT), |ctx| {
        FieldFuture::new(async move {
            println!("args  {:?}", ctx.args.as_index_map());
            Ok(Some(Value::from(10)))
        })
    })
    .argument(InputValue::new("q", TypeRef::named("Int")).default_value(value!(12)))
);

tokio::runtime::Runtime::new().unwrap().block_on(async move {

let schema = Schema::build(query.type_name(), None, None)
    .register(query)
    .finish()?;

let val =  schema
.execute("{ test(q: 12) }")
.await;
println!("val : {:?} ", val);

```

In this example, the field test has an argument q with a default value of 12. When the argument is provided explicitly, everything works as expected. However, when the argument is omitted, the default value is not applied, leading to unexpected behavior.

### Steps to Reproduce

- Define a field with an argument that has a default value.
- Execute a query without providing the argument.
- Observe that the default value is not applied.

### Expected Behaviour

When the argument q is omitted from the query, the default value 12 should be used.

### Actual Behaviour

When the argument q is omitted, the default value is not applied, and the query execution does not behave as expected.

Example Code

Here is the code snippet that illustrates the issue:
```graphql
use async_graphql::{dynamic::*, value, Value};

let query = Object::new("Query").field(
    Field::new("test", TypeRef::named_nn(TypeRef::INT), |ctx| {
        FieldFuture::new(async move {
            println!("args  {:?}", ctx.args.as_index_map());
            Ok(Some(Value::from(10)))
        })
    })
    .argument(InputValue::new("q", TypeRef::named("Int")).default_value(value!(12)))
);

tokio::runtime::Runtime::new().unwrap().block_on(async move {

let schema = Schema::build(query.type_name(), None, None)
    .register(query)
    .finish()?;

let val =  schema
.execute("{ test }")  // Argument q is omitted here
.await;
println!("val : {:?} ", val);

```

ssddOnTop:
Just visited this repo to raise the same issue. Dropping this comment to get participated automatically

PR #1527, opened 2024-05-25
fix: pass default vals to `ResolverContext`

ssddOnTop:
**Issue Reference**
fixes: #1526

**Performace issue**
This will reduce performance because of merging maps and clones. Additionally, it is a hot code so I don't know if this solution is good enough.


ssddOnTop:
Also, I am not sure where to add tests.. so please guide me a little

ssddOnTop:
@sunli829 can I get a review pls

sunli829:
thanks

PR #1530, opened 2024-06-01
Add altair source

lasantosr:
Added new [altair](https://github.com/altair-graphql/altair) source, as an optional replacement for GraphiQL.

 Async-graphql have support for file uploads but GraphiQL doesn't, so I've implemented altair also under a non-default feature flag to optionally get it.

The configuration can be provided using _json!_ macro or the strongly typed struct, which has support for deserialization and schemars in case anyone needs to store it on a file or dynamically set it up.


sunli829:
thanks, altair looks great. 🙂

PR #1534, opened 2024-06-06
feat: Add support for using `Interface` and `OneofObject` on the same struct

PookieBuns:
Currently there is support for defining `Union` and `OneofObject` on the same struct thanks to https://github.com/async-graphql/async-graphql/commit/7f70c7e765682ea1c31b1d1d68650695f2775af7

This PR aims to support defining `Interface` and `OneofObject`.

Note: This doesn't enable support for using user-defined resolvers for the fields. The interface fields must exist on the Object struct


sunli829:
looks useful 😂

PR #1535, opened 2024-06-08
feat: Allow `skip` and `derived` together to support proxy types.

### LINKED ISSUE #1490: As using proxy type
## Description of the feature
I would like to include a feature similar to [`serde_with::FromInto`](https://docs.rs/serde_with/3.7.0/serde_with/struct.FromInto.html): 
> Serialize value by converting to/from a proxy type with serde support.
>
> This adapter serializes a type O by converting it into a second type T and serializing T. Deserializing works analogue, by deserializing a > T and then converting into O.

```rs
struct S {
    #[serde_as(as = "FromInto<T>")]
    #[graphql(as = "T")]
    value: O,
}
```

The latter eliminates the requirement for wrapper types and/or additional logic because it uses something that GraphQL already understands (i.e. `u8`, `u16`, `String`, ...).

Additionally, it can be used in conjunction with `skip_input` to eliminate it from an `InputObject`.

## Code example
```rs
#[serde_as]
#[derive(Serialize, Deserialize, SimpleObject, InputObject, Copy, Clone, Debug, PartialEq, Eq, Hash)]
pub struct MyObject {
    #[serde_as(as = "FromInto<u8>")]
    #[graphql(as = "u8")]
    pub type: external::enum::Type, // Assume u8 via impl From<u8> and impl From<external::enum::Type>
    #[serde_as(as = "FromInto<u16>")]
    #[serde_as(skip_input, as = "u16")]
    pub only_object: external::Obj, // Assume u16 (only Object) via impl From<external::Obj> for u16
    pub name: String,
}
```

- `type` is `external::enum::Type`, however it is seen as an `u8` in both `SimpleObject` and `InputObject` thanks to `impl From<u8> for external::enum::Type` and `impl From<external::enum::Type> for u8`
- `only_object` is `external::Obj`, however it is seen as an `u16` in `SimpleObject` thanks to `impl From<external::enum::Type> for u16`
- `only_object` is skipped in `InputObject` thanks to `skip_input`

```graphql
query {
  myObj {
    type, # 100
    onlyObject, # 65535
    name # "Hello World!"
  }
}
```

PookieBuns:
Right now, when the field skip is enabled, derived fields all get ignored. This change aims to allow derived fields while using skip simultatenously to be able to use a proxy type for graphql queries.

Partially Fixes #1490

Note:
- Does not provide support for complex object (IMHO I don't really see a use case for using skip on user defined resolvers since they are graphql exclusive don't have to be compatible with other structs per se)
- Does not provide support for InputObject

### DIFF
--- a/derive/src/simple_object.rs
+++ b/derive/src/simple_object.rs
@@ -114,7 +114,7 @@ pub fn generate(object_args: &args::SimpleObject) -> GeneratorResult<TokenStream
     }
 
     for SimpleObjectFieldGenerator { field, derived } in &processed_fields {
-        if field.skip || field.skip_output {
+        if (field.skip || field.skip_output) && derived.is_none() {
             continue;
         }
 
--- a/tests/derived_field.rs
+++ b/tests/derived_field.rs
@@ -382,3 +382,57 @@ pub async fn test_derived_field_complex_object_derived() {
         })
     );
 }
+
+#[tokio::test]
+pub async fn test_derived_field_with_skip_simple_object() {
+    use serde::{Deserialize, Serialize};
+
+    struct Query;
+
+    #[derive(Serialize, Deserialize)]
+    struct ValueDerived(String);
+
+    scalar!(ValueDerived);
+
+    impl From<i32> for ValueDerived {
+        fn from(value: i32) -> Self {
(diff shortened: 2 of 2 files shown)


sunli829:
thanks! 🙂

Issue #241, opened 2020-08-14
Interface as an object and union


Koxiaet:
Support interfaces that are objects and a union, to avoid repeating the interface fields for each implementation. Example:

```rust
#[Interface]
struct Character {
    name: String,
    __implementations: CharacterTypes,
}

#[Union]
enum CharacterTypes {
    Human(Human),
    Droid(Droid),
}
```

Or, with fields:

```rust
#[Interface]
impl Shape {
    async fn area(&self) -> f32 { ... }
    async fn scale(&self, s: f32) -> Self { ... }
    async fn __implementations(&self) -> CharacterTypes { ... }
}
```
`__implementations` could also be called `__impls` because it's shorter.

sunli829:
@Koxiaet @chipsenkbeil 

how about this? 🙂

```rust
#[derive(Interface)]
#[graphql(impl)]
enum CharacterTypes {
    Human(Human),
    Droid(Droid),
}

#[InterfaceImpl]
impl CharacterTypes {
    async fn id(&self) -> &str {
        todo!()
    }

    async fn name(&self) -> &str {
        todo!()
    }
}
```

sunli829:
This cannot be verified at compile time. 🥺

> The object type must include a field of the same name for every field defined in an interface.

Koxiaet:
The idea is that the outer struct provides all the common fields - the inner interface/union doesn't know about any fields of the outer one. It's for the case when the shared fields of the interface are resolved under common logic - so instead of being repeated for each variant in the interface, the code only has to be written once.

So ideally it would require no changes to the definition of the inner interface/union - the outer interface will simply add more fields to the inner one.

sunli829:
Yes, so I'm still thinking about how to add fields to the objects inside.

chipsenkbeil:
@sunli829 did you have any new thoughts about how to do this? I've got a trait that all of the variants in my union implement and it would be a lot easier to reference those trait methods like this

```graphql
query AllTheData {
    getSomeDataAsList {
        id
        type
        ... on Paragraph {
            text
        }
    }
}
```

instead of this

```graphql
query AllTheData {
    getSomeDataAsList {
        ... on Header {
            id
            type
        }
        ... on Section {
            id
            type
        }
        ... on Picture {
            id
            type
        }
        ... on Paragraph {
            id
            type
            text
        }
    }
}
```

sunli829:
[...]

PR #1539, opened 2024-06-13
feat: Enum implementations for `Interface` + Support Unit Structs for `Interface`

PookieBuns:
## Enum implementations for `Interface`

I took a jab at #241 

this feature allows using `ComplexObject(interface)` to define enum level implementations of an interface field. An example is shown below
```rs
#[derive(SimpleObject, Clone)]
#[graphql(interface_impl(interface_type = "Node", name = "id", ty = "i32"))]
struct NewType {
    title_a: String,
}

#[derive(SimpleObject, Clone)]
#[graphql(interface_impl(interface_type = "Node", name = "id", ty = "i32"))]
struct NewTypeUnit;

#[derive(Interface)]
#[graphql(field(name = "id", ty = "i32"))]
enum Node {
    NewType(NewType),
    NewTypeUnit(NewTypeUnit),
}

#[ComplexObject(interface)]
impl Node {
    async fn id(&self) -> i32 {
        match self {
            Node::NewType(_) => 1,
            Node::NewTypeUnit(_) => 2,
        }
    }
}
```
The attribute macro will call convert the newtype variant into its parent enum and use it to resolve the field.

note:
- annotating with interface_impl will introduce a new method called interface_impl_{field} so it doesn't collide with the function signature expanded from `graphql(field)`
- Clone is required on structs that use interface_impl in order to perform type conversions
- My initial goal was to be able to implement this without having to do anything to the child newtype structs. However, I found this quite difficult since attribute macros on the enum could only modify the resolver for the enum but not its child types. This would create a logical error when querying in that if the newtype also defined a field with the same name, querying like `query { id }` would return the enum resolver result and querying like `query { ... on NewType { id } }` would return the newtype struct resolver result. So I basically concluded the most pragmatic way without huge changes was to use an attribute macro to change the newtype resolver and still let the enum propagate to the child resolver for consistency
- This could be a breaking change because I added an implementation function to the ComplexObject trait. Although this trait is hidden in the docs, a user could still manually implement this trait for their use case.

## Support Unit Structs for `Interface`
With the feature above implemented, we can now support unit variants of enums. An example is shown below
```rs
#[derive(Interface)]
#[graphql(field(name = "id", ty = "i32"))]
enum Node {
    NewType(NewType),
    NewTypeUnit(NewTypeUnit),
    Unit
}

#[ComplexObject(interface)]
impl Node {
    async fn id(&self) -> i32 {
        match self {
            Node::NewTypeUnit(_) => 2,
            Node::Unit => 3,
            // This will get ignored!
            Node::NewType(_) => 999,
        }
    }
}

#[derive(SimpleObject)]
struct NewType {
    #[graphql(owned)]
    id: i32,
}

#[derive(SimpleObject, Clone)]
#[graphql(interface_impl(interface_type = "Node", name = "id", ty = "i32"))]
struct NewTypeUnit;
```

On macro expansion, a hidden struct `Unit` will get defined and will utilize the Enum implementation for resolving the interface field.

---

Feel free to let me know what you think about this approach and / or suggestions / comments! Would love to explore other ways we can improve the user experience for this library!

### DIFF
--- a/derive/src/args.rs
+++ b/derive/src/args.rs
@@ -232,6 +232,13 @@ pub struct SimpleObjectField {
     pub directives: Vec<Expr>,
 }
 
+#[derive(FromMeta)]
+pub struct InterfaceImpl {
+    pub interface_type: Type,
+    pub name: SpannedValue<String>,
+    pub ty: Type,
+}
+
 #[derive(FromDeriveInput)]
 #[darling(attributes(graphql), forward_attrs(doc))]
 pub struct SimpleObject {
@@ -266,6 +273,8 @@ pub struct SimpleObject {
     pub interface_object: bool,
     #[darling(default, multiple, rename = "tag")]
     pub tags: Vec<String>,
+    #[darling(default, multiple, rename = "interface_impl")]
+    pub interface_impls: Vec<InterfaceImpl>,
     #[darling(default)]
     pub visible: Option<Visible>,
     #[darling(default, multiple, rename = "concrete")]
@@ -940,6 +949,7 @@ pub struct ComplexObject {
     pub rename_fields: Option<RenameRule>,
     pub rename_args: Option<RenameRule>,
     pub guard: Option<Expr>,
+    pub interface: bool,
 }
 
 #[derive(FromMeta, Default)]
--- a/derive/src/args.rs
+++ b/derive/src/args.rs
@@ -237,6 +237,9 @@ pub struct InterfaceImpl {
     pub interface_type: Type,
     pub name: SpannedValue<String>,
     pub ty: Type,
+
+    #[darling(default)]
+    pub method: Option<String>,
 }
 
 #[derive(FromDeriveInput)]
--- a/derive/src/complex_object.rs
+++ b/derive/src/complex_object.rs
@@ -171,7 +171,7 @@ pub fn generate(
             });
             if object_args.interface {
                 method.sig.ident = Ident::new(
-                    &format!("{}_interface", method.sig.ident),
+                    &format!("interface_impl_{}", method.sig.ident),
                     method.sig.ident.span(),
                 );
             }
--- a/derive/src/complex_object.rs
+++ b/derive/src/complex_object.rs
@@ -30,6 +30,7 @@ pub fn generate(
 
     let mut resolvers = Vec::new();
     let mut schema_fields = Vec::new();
+    let mut register_fields = Vec::new();
 
     // Computation of the derivated fields
     let mut derived_impls = vec![];
@@ -168,6 +169,12 @@ pub fn generate(
                     .rename_fields
                     .rename(method.sig.ident.unraw().to_string(), RenameTarget::Field)
             });
+            if object_args.interface {
+                method.sig.ident = Ident::new(
+                    &format!("{}_interface", method.sig.ident),
+                    method.sig.ident.span(),
+                );
+            }
             let field_desc = get_rustdoc(&method.attrs)?
                 .map(|s| quote! { ::std::option::Option::Some(::std::string::ToString::to_string(#s)) })
                 .unwrap_or_else(|| quote! {::std::option::Option::None});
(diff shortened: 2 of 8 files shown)


sunli829:
Thank you very much for your contribution. 🙂 Regarding this PR, I have the following questions:

1. What is the benefit of `#[ComplexObject(interface)]`? It seems more reasonable to have the resolver of the field on each subtype?

2. It is more troublesome to implement an interface now than before. Each field must be declared with `#[graphql(interface_impl(..))]` on each subtype.

3. Each subtype must be cloneable, which limits the use of this feature.

PookieBuns:
1. From a user's perspective, I would like to have common fields that can be implemented using one function at the enum level. In it's most ideal state, it would look like this
> @sunli829 from a library user's perspective, something like this:
> 
> ```rust
> use async_graphql::*;
> 
> #[derive(Union)]
> #[graphql(complex)] // NOTE: If you want the `ComplexObject` macro to take effect, this `complex` attribute is required.
> enum Shape {
>     Circle(Circle),
>     Square(Square),
> }
> 
> 
> #[ComplexUnion]
> impl Shape {
>     async fn area(&self) -> f32 {
>         match self {
>             Self::Circle(x) => std::f32::consts::PI * x.radius * x.radius,
>             Self::Square(x) => x.width * x.width,
>         }
>         
>     }
> 
>     async fn scale(&self, s: f32) -> Shape {
>         match self {
>             Self::Circle(x) => Circle { radius: x.radius * s }.into(),
>             Self::Square(x) => Square { width: x.width * s }.into()
>         }
>     }
> }
> 
> struct Circle {
>     radius: f32,
> }
> 
> #[Object]
> impl Circle {
>     async fn radius(&self) -> f32 {
>         self.radius
>     }
> }
> 
> struct Square {
>     width: f32,
> }
> 
> #[Object]
> impl Square {
>     async fn width(&self) -> f32 {
>         self.width
>     }
> }
> ```
> 
> enables a graphql query like
> 
> ```graphql
> {
>     some_shape(...) {
>         area
>         scale(s: 3.0) {
>             area
>         }
>         ... on Circle {
>             radius
>         }
>         ... on Square {
>             width
>         }
>     }
> }
> ```
> 
> This follows the interface of graphql (instead of union), but enables implementing all of the common fields in a single place instead of needing to repeat them for every single object:
> 
> image
Currently, if I wanted to do something similar, I would have to do something like this
```rs
use async_graphql::*;

#[derive(Interface)]
#[graphql(field(name = "area", ty = "f32"))]
enum Shape {
    Circle(Circle),
    Square(Square),
}

struct Circle {
    radius: f32,
}

#[Object]
impl Circle {
    async fn area(&self) -> f32 {
        std::f32::consts::PI * self.radius * self.radius
    }

    async fn radius(&self) -> f32 {
        self.radius
    }
}

struct Square {
    width: f32,
}

#[Object]
impl Square {
    async fn area(&self) -> f32 {
        self.width * self.width
    }

    async fn width(&self) -> f32 {
        self.width
    }
}
```
This requires manual implementation of area on each struct which could lead to quite a bit of duplication and non-ideal code structure. On the other hand, if I wanted to let the enum resolve this field, I could do something like this
```rs
#[derive(Interface)]
#[graphql(field(name = "area", ty = "f32"))]
enum Shape {
    Circle(Circle),
    Square(Square),
}

impl Shape {
    fn my_area(&self) -> f32 {
        match self {
            Self::Circle(x) => std::f32::consts::PI * x.radius * x.radius,
            Self::Square(x) => x.width * x.width,
        }
        
    }
}

#[derive(Clone)]
struct Circle {
    radius: f32,
}

#[Object]
impl Circle {
    async fn area(&self) -> f32 {
        Shape::from(self.clone()).my_area()
    }

    async fn radius(&self) -> f32 {
        self.radius
    }
}

#[derive(Clone)]
struct Square {
    width: f32,
}

#[Object]
impl Square {
    async fn area(&self) -> f32 {
        Shape::from(self.clone()).my_area()
    }
    async fn width(&self) -> f32 {
        self.width
    }
}
```
Both of these methods don't seem ideal since they require manually implementing the field resolver when it theoretically can be handled by the enum.

However, during my implementation trying to achieve the ideal case, I realized the difficulty since there is no way of intercepting the newtype struct to modify its implementation of output type with only a macro on the enum (without breaking graphql spec). That's why I approached the next best thing which is allow users to use an attribute macro to use its parent method instead of having to manually implement the resolver by itself.

## The big advantage to allowing enum implementations is that now we can support unit structs variants for OutputType enums. This gives users more flexibility to write something simple like
```rs
use async_graphql::*;

#[derive(Interface)]
#[graphql(field(name = "area", ty = "f32"))]
enum Shape {
    Circle,
    Square(Square),
}

#[ComplexObject(interface)]
impl Shape {
    async fn area(&self) -> f32 {
        match self {
            Self::Circle => std::f32::consts::PI,
            // ignored
            _ => 0.0,
        }
        
    }
}

#[derive(Clone)]
struct Square {
    width: f32,
}

#[Object]
impl Square {
    async fn area(&self) -> f32 {
        self.width * self.width
    }
    async fn width(&self) -> f32 {
        self.width
    }
}
```

2. This feature is opt in, since you could still manually define the field in your struct instead of using the enum implementation like show above

3. That is true. However in the current version of this crate, if I wanted to use an enum resolver I would still need to clone the object.

I'll admit the need to annotate ```interface_impl``` is non ideal and still requires some duplication, but I think the ability to use unit structs on interface enums would be really nice. The feature being opt in also maintains backwards compatibility for using interface (Could still break complex object).

If we really don't like the annotation on SimpleObject, what do you think of hiding that feature and only exposing ```ComplexObject(interface)``` to allow unit variant enums?

sunli829:
Sorry for the late reply.
I admit that this feature is somewhat useful in some cases, but in most cases, it does not simplify the code.

PookieBuns:
Thanks for the feedback! I guess my use case was perhaps too specific or I haven't found the ideal solution to it. Will look for another method to support this some other time 😄

sunli829:
thank you very much for your efforts. 🙂

PR #1548, opened 2024-06-24
feat: get directives from SelectionField

### LINKED ISSUE #1547: feat: Inspect query directives from SelectionField
## Description of the feature

Our use case is to proxy graphql requests to other graphql backends. To implement this we use `SelectionField` to reconstruct graphql query specific to every backend. But for now SelectionField is only providing info about the name, alias and arguments. In case the request defines any field directives there is no way to get it.

The issue is about providing access to query directives from the `SelectionField`. Seems like it could be done by `field` property of `SelectionField` itself, but since it's a private field probably special method should be added.

meskill:
fixes #1547

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -8,6 +8,7 @@ use std::{
     sync::{Arc, Mutex},
 };
 
+use async_graphql_parser::types::ConstDirective;
 use async_graphql_value::{Value as InputValue, Variables};
 use fnv::FnvHashMap;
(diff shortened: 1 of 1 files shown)


meskill:
@sunli829 could you please review this?

ouertani:
Any update on this PR please?

sunli829:
thanks 🙂

PR #1551, opened 2024-06-25
fix: handle the case where the response contains null for the enum 

ouertani:
Should handle the case where the response contains null for the enum


xamgore:
Why?

ouertani:
@xamgore I got error when enum from the graphql response is `null`

xamgore:
Could you show the case, please? To understand, either it is a problem in the implementation and needs to be fixed, or just a problem in your source code.

Akremkadri (approved):

ouertani:
@xamgore I tired to reproduce with the vanilla `async-graphql` crate and I confirm this is in issue in the related library and the source code I'm referring too. I will close this PR and thank you for the support.

PR #1554, opened 2024-07-01
Support raw values from serde_json

IvanUkhov:
The motivation here is the same as the one for `raw_value` in `serde_json`:

https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html

Namely, in case of large payloads that do not have to be processed but just forwarded further, it becomes very CPU and memory intensive to have them parsed into `ConstValue` and then serialized back to JSON. Please see the test for an illustration.

### DIFF
--- a/value/src/lib.rs
+++ b/value/src/lib.rs
@@ -27,6 +27,9 @@ pub use serde_json::Number;
 pub use serializer::{to_value, SerializerError};
 pub use variables::Variables;
 
+#[cfg(feature = "raw_value")]
+pub use value_serde::RAW_VALUE_TOKEN;
+
 /// A GraphQL name.
 ///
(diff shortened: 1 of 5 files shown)


value/src/lib.rs
```diff
@@ -27,6 +27,9 @@ pub use serde_json::Number;
 pub use serializer::{to_value, SerializerError};
 pub use variables::Variables;
 
+#[cfg(feature = "raw_value")]
+pub use value_serde::RAW_VALUE_TOKEN;
```
IvanUkhov, line 31: Does not have to be exposed, but it is convenient to have it exposed still. `serde_json` should arguably do the same.

sunli829:
thanks 🙂

PR #1563, opened 2024-07-20
chore: Make Extensions nullable

ssddOnTop:
Currently `extensions` in `GraphQLRequest` are just HashMap of String, Value, so for the sample query:

```gql
{
   "query": "{ foo }",
   "extensions": null
}
```

deserialization will fail because of the null value

to avoid that, I added the `Extensions` struct which handles deserialization properly with minimal change.


ssddOnTop:
@sunli829 simple change.. can I get a quick review

sunli829:
thanks 🙂

PR #1564, opened 2024-07-20
chore: skip serializing if Response data is default

ssddOnTop:
Currently response data is serialized to `null` but according to GraphQL Over HTTP rules, it should be skipped.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -9,7 +9,7 @@ use crate::{CacheControl, Result, ServerError, Value};
 #[derive(Debug, Default, Serialize, Deserialize, PartialEq)]
 pub struct Response {
     /// Data of query result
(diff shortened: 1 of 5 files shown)


ssddOnTop:
@Miaxos @dbanty @sunli829 it's been a while.. can I get a review on this

dbanty:
According to the GraphQL spec, there is a semantic difference between `data: null` and omitting `data`.

If execution has started, data should be null, not missing. Only if the request failed before execution should data be omitted.

http://spec.graphql.org/October2021/#sec-Response-Format

So we can't always skip serializing null, only when there was a fundamental error (which would also result in a 4xx or 5xx status according to graphql-over-http)

PR #1565, opened 2024-07-23
Add an "on ping" callback

bakhtiyarneyman:
This PR works but probaby is not ready to be merged. Sending this PR out to see if @sunli829 thinks this is the right approach (it's very bare-bones). 

Our application needs to react to subscription disconnections. This PR adds an `on_ping` callback which can help with that. We rely on the client sending pings at some fixed interval. In the `on_ping` callback the server could repeatedly restart a timer, upon expiry of which the server could declare the connection to be dead and do something. 

### Rough edges

- The `on_ping` callback is NOT an async function, but it is invoked by the loop in `serve`. This means it's a potential footgun. If the user accidentally blocks here, the subscription might not react in time to necessary events. In practice, most of the users would probably invoke a `send` method on `broadcast::Sender` from that call. The corresponding `Receiver` needs to be passed as data to the context, and thus accessed by the subscription stream.
- Relying on the user sending pings at a certain interval (as opposed to server sending the pings and awaiting the pongs) requires coordination between the client and the server on what that ping duration is. This approach was chosen for its simplicity.

If this approach is approved I can add tests and extend other integrations. 

##### Async callback 

We could have been accepting an async function as a callback and calling `tokio::spawn`. Rejected because introduces a dependency on the `tokio` crate and is not zero-cost. 

##### Ping stream

Instead of an `on_ping` callback, we could have allowed the user to supply a `Stream` that produces pings, and an `on_pong` callback. Rejected because it's more complex.

### DIFF
--- a/integrations/axum/src/subscription.rs
+++ b/integrations/axum/src/subscription.rs
@@ -2,7 +2,7 @@ use std::{borrow::Cow, convert::Infallible, future::Future, str::FromStr};
 
 use async_graphql::{
     futures_util::task::{Context, Poll},
-    http::{WebSocketProtocols, WsMessage, ALL_WEBSOCKET_PROTOCOLS},
+    http::{DefaultOnPingType, WebSocketProtocols, WsMessage, ALL_WEBSOCKET_PROTOCOLS},
     Data, Executor, Result,
 };
 use axum::{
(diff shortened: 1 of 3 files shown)


sunli829:
I've added `Websocket::keepalive_timeout` method to sets a timeout for receiving an acknowledgement of the keep-alive ping. If the ping is not acknowledged within the timeout, the connection will be closed.

https://github.com/async-graphql/async-graphql/commit/ae589dc7ffdb06ea9e5b8d0ff1d50425a7b70fea

bakhtiyarneyman:
Hi Sunli! Thank you for the update! My two cents:

This is much nicer than what we have in terms of ergonomics, but I _think_ your approach is not taking into consideration the GraphQL spec regarding payload on pings. Our use-case is a chatroom, and we use ping payloads to deliver ephemeral information, such as whether the user is typing a message. Is there something that I'm missing here?

sunli829:
added `on_ping` callback to `WebSocket` 🙂

https://github.com/async-graphql/async-graphql/commit/da5f8db83c7dc29d0e3724f03ffc3b0eab138118

sunli829:
released in `v7.0.9`

bakhtiyarneyman:
Awesome, thank you, Sunli! 

This is a minor point, I don't quite understand the need to pass the Data in context. In practice, the user will have the data in context when building GraphQLWebSocket, because they have to call `with_data`. 

E.g. 
```
      upgrade
        .protocols(async_graphql::http::ALL_WEBSOCKET_PROTOCOLS)
        .on_upgrade(move |socket| {
          async_graphql_axum::GraphQLWebSocket::new(socket, schema, protocol)
            .with_data(data)
            .on_ping({
              move |_, payload| async move {
                ...
              }
            })
            .serve()

```

Maybe it's for some other way to set up subscriptions that I'm not aware of.

PR #1566, opened 2024-07-23
Bump opentelemetry 0.23 > 0.24

robmarto:
Bump otel to 0.24 which is using http 1.0. No code changes required.


b-dreissigacker:
@Miaxos @sunli829  any plans of merging this soon? I'm stuck on older versions of other libraries because async-graphql breaks if I update otel to 0.24 in my project.

fiadliel:
@b-dreissigacker Apparently a new otel release is planned for this week, so another trip on the upgrade merry-go-round. IMO otel dependencies (in all projects, not just this one) should be split out into a separate crate; the API is cross-cutting, and too unstable right now.

PR #1569, opened 2024-08-02
fix: extensions should be optional

karatakis:
According to graphql-over-http the extension option should be optional. I adjusted the code to mitigate this issue.

> extensions - (Optional, map): This entry is reserved for implementors to extend the protocol however they see fit.

== Source ==
https://graphql.github.io/graphql-over-http/draft/#sel-FALFFHDAAACD8Cokd


ssddOnTop:
duplicate of https://github.com/async-graphql/async-graphql/pull/1563

sunli829:
fixed in #1563

karatakis:
thanks @sunli829

PR #1581, opened 2024-08-22
Support scraping `#[doc = ...]` attributes when generating descriptions

parrottq:
This modifies the Rust doc scraping function used by all `async-graphql` macros to start scrape `#[doc = ...]` attributes and adding them to the GraphQL description of the related type. This is particularly useful for including external documentation using an attribute like `#[doc = include_str!("external_doc.md")]`. Previously `#[doc]` attributes were ignored. Using the `#[doc]` attribute should be supported by all `async-graphql` macros after this change as far as I can tell.

Example:
```rust
use async_graphql::*;

struct MyObject;

#[Object]
impl MyObject {
    /// Return a value
    ///
    #[doc = include_str!("external_doc.md")] // <--- This will be included in the GraphQL description.
    ///
    /// You can also include text down here.
    async fn value(&self) -> String {
        "Some text".to_string()
    }
}
```

### DIFF
--- a/derive/src/utils.rs
+++ b/derive/src/utils.rs
@@ -71,9 +71,7 @@ pub fn get_rustdoc(attrs: &[Attribute]) -> GeneratorResult<Option<TokenStream>> 
                     }) => {
                         let doc = doc.value();
                         let doc_str = doc.trim();
-                        if !combined_docs_literal.is_empty() {
-                            combined_docs_literal += "\n";
-                        }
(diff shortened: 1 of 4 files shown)


sunli829:
Could you add a test case? 🙂

parrottq:
Added a few test cases that I think will cover the most important regressions. Feel free to tell me if you spot anything missing.

sunli829:
thanks, that's good enough.

PR #1582, opened 2024-08-24
Impl Debug for dynamic::FieldValue & Improve error messages for its methods

hozan23:
Hello,

This PR implements the `Debug` trait for the `dynamic::FieldValue` type, which enhances the ability to debug issues

**Context:**

Consider the following code snippet, which contains a trivial bug where an `Option<Book>` is used instead of a `Book` for the `FieldValue`:

```rust
let book_query_field = Field::new("book", TypeRef::named(book.type_name()), |ctx| {
    FieldFuture::new(async move {
        let id = ctx.args.try_get("id")?;
        let book_id = match id.string() {
            Ok(id) => id.to_string(),
            Err(_) => id.u64()?.to_string(),
        };
        let book_id = book_id.parse::<usize>()?;
        let store = ctx.data_unchecked::<Storage>().books.lock().await;
        let book = store.get(book_id).cloned();
        Ok(Some(FieldValue::owned_any(book)))
    })
})
.argument(InputValue::new("id", TypeRef::named_nn(TypeRef::ID)));
```

This code results in an ambiguous error:

```
internal: not type Book
```

Without the `Debug` trait implemented, it’s challenging to identify the root cause of this issue at first glance. 

This PR is also Improve error messages in FieldValue methods for better debugging clarity.

### DIFF
--- a/src/dynamic/field.rs
+++ b/src/dynamic/field.rs
@@ -21,9 +21,11 @@ pub(crate) enum FieldValueInner<'a> {
     /// Const value
     Value(Value),
     /// Borrowed any value
-    BorrowedAny(&'a (dyn Any + Send + Sync)),
+    /// The first item is the [`std::any::type_name`] of the value used for debugging.
+    BorrowedAny(Cow<'static, str>, &'a (dyn Any + Send + Sync)),
     /// Owned any value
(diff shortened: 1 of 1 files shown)


sunli829:
thanks 🙂

PR #1588, opened 2024-09-07
UUID validator

edbo:
Add a validator that accepts ID or String Input Values that are UUIDs and validates them for correctness and checks the version number is correct.

Example:

```
#[Object]
impl MutationRoot {
    async fn add_user_to_team(
        &self,
        ctx: &Context<'_>,
        #[graphql(validator(uuid = 4))] team_id: ID,
        #[graphql(validator(uuid = 4))] user_id: ID,
    ) -> ID {
        let pool = ctx.data::<Pool<Postgres>>().unwrap();
        team_repository::add_user_to_team(
            &pool,
            Uuid::parse_str(&team_id).unwrap(),
            Uuid::parse_str(&user_id).unwrap(),
        )
        .await
        .unwrap();
        team_id
    }
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ graphiql = ["handlebars"]
 altair = ["handlebars", "schemars"]
 playground = []
 raw_value = ["async-graphql-value/raw_value"]
+uuid-validator = ["uuid"]
 
 [[bench]]
 harness = false
--- a/derive/src/validators.rs
+++ b/derive/src/validators.rs
@@ -58,6 +58,8 @@ pub struct Validators {
     ip: bool,
     #[darling(default)]
     regex: Option<String>,
+    #[darling(default)]
+    uuid: Option<usize>,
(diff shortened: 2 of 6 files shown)


sunli829:
thanks 🙂

edbo:
I went with 

```
#[graphql(validator(uuid))] team_id: ID,
#[graphql(validator(uuid_version = 4))] user_id: ID,
```

I couldn't find a simple way to have a proc arg accept a bool or an int but open to suggestions if there is a way.

sunli829:
Do it like this

https://github.com/async-graphql/async-graphql/blob/89b48e2a8d7cbccdb7440e233031534c9ac3ac46/derive/src/args.rs#L49

edbo:
@sunli829 Sorry for the delay and thanks so much for the pointer. Not sure why I didn't try just putting a type in the option.

I've rebased to master so hopefully this is good to go now.

Issue #1592, opened 2024-09-13
 cargo clippy --all failing


jayvdb:
https://github.com/async-graphql/async-graphql/actions/runs/10750783522/job/29839224992?pr=1587

```
$ cargo clippy --all
error: `?Sized` bound is ignored because of a `Sized` requirement
   --> src/registry/mod.rs:915:24
    |
915 |         T: InputType + ?Sized,
    |                        ^^^^^^
    |
note: `T` cannot be unsized because of the bound
   --> src/registry/mod.rs:915:12
    |
915 |         T: InputType + ?Sized,
    |            ^^^^^^^^^
    = note: ...because `InputType` has the bound `Sized`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_maybe_sized
note: the lint level is defined here
   --> src/lib.rs:140:9
    |
140 | #![deny(clippy::all)]
    |         ^^^^^^^^^^^
    = note: `#[deny(clippy::needless_maybe_sized)]` implied by `#[deny(clippy::all)]`
help: change the bounds that require `Sized`, or remove the `?Sized` bound
    |
915 -         T: InputType + ?Sized,
915 +         T: InputType,
    |

error: could not compile `async-graphql` (lib) due to 1 previous error
```

PR #1593, opened 2024-09-13
remove needless ?Sized

jayvdb:
Closes https://github.com/async-graphql/async-graphql/issues/1592

Also set the MSRV to 1.75.
1.74 fails with
```
> cargo +1.74 build
   Compiling async-graphql v7.0.9 (/home/jayvdb/rust/async-graphql)
error[E0562]: `impl Trait` only allowed in function and inherent method return types, not in trait method return types
  --> src/base.rs:92:10
   |
92 |     ) -> impl Future<Output = ServerResult<Value>> + Send;
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #91611 <https://github.com/rust-lang/rust/issues/91611> for more information

....
error: could not compile `async-graphql` (lib) due to 91 previous errors; 2 warnings emitted
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ categories = ["network-programming", "asynchronous"]
 description = "A GraphQL server library implemented in Rust"
 documentation = "https://docs.rs/async-graphql/"
 edition = "2021"
+rust-version = "1.75"
 homepage = "https://github.com/async-graphql/async-graphql"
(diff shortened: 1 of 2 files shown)


sunli829:
thanks 🙂

PR #1602, opened 2024-09-25
connection: Allow 'first' and 'last' parameters to exist at the same time

### LINKED ISSUE #1601: query(): The "first" and "last" parameters should be allowed to exist at the same time
## Expected Behavior
In [`async_graphql::types::connection::query`](https://docs.rs/async-graphql/7.0.10/async_graphql/types/connection/fn.query.html) function, the "first" and "last" parameters should be allowed to exist at the same time, as per [GraphQL Cursor Connection Specification](https://relay.dev/graphql/connections.htm#sec-Pagination-algorithm).

## Actual Behavior
The "first" and "last" parameters cannot exist at the same time.

https://github.com/async-graphql/async-graphql/blob/04558c6b252d036e064c3461a0b72f6d2b930eac/src/types/connection/mod.rs#L355-L357

simnalamburt:
The "first" and "last" parameters should be allowed to exist at the same time, as per [GraphQL Cursor Connection specification](https://relay.dev/graphql/connections.htm#sec-Pagination-algorithm).

Reference: https://relay.dev/graphql/connections.htm#sec-Pagination-algorithm

Closes #1601

### DIFF
--- a/src/types/connection/mod.rs
+++ b/src/types/connection/mod.rs
@@ -352,10 +352,6 @@ where
     R: Future<Output = Result<T, E>>,
     E: Into<Error>,
 {
-    if first.is_some() && last.is_some() {
-        return Err("The \"first\" and \"last\" parameters cannot exist at the same time".into());
-    }
-
     let first = match first {
         Some(first) if first < 0 => {
             return Err("The \"first\" parameter must be a non-negative number".into());


Sytten:
This is kinda bad that it is in a patch since a lot of people rely on that...

PR #1609, opened 2024-10-03
Collect extra fields from the request

XiNiHa:
This lets extensions receive and convert nonstandard requests 😋

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,5 +1,6 @@
 use std::{
     any::Any,
+    collections::HashMap,
     fmt::{self, Debug, Formatter},
 };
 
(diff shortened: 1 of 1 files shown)


sunli829:
you can insert extra fields into `Request.extensions`

XiNiHa:
Wouldn't that require putting extra fields into `extensions`, rather than putting it on top level? Although that could be sufficient for some cases, it wouldn't work if the field should be placed on top level due to spec requirements.

XiNiHa:
Well, I wanted to make it simple to integrate with extensions, rather than including additional steps to collect extra fields ;( I understand your intention though 👍

PR #1616, opened 2024-10-09
Deprecate `data_unchecked` and `data_opt` methods

patrickariel:
There are currently three ways of retrieving data:

- `data()` -> returns `Result<D>`
- `data_unchecked()` -> panicking version of `data()`
- `data_opt()` -> like `data()`, but returns `Option<D>`

But the API feels awkward and unidiomatic for a few reasons:

1. `data_unchecked()` is a hugely misleading name. `*_unchecked` is a name usually reserved for unsafe functions that can potentially invoke undefined behavior (e.g. [`get_unchecked`](https://doc.rust-lang.org/std/primitive.slice.html#method.get_unchecked)). Here, it's simply a panicking version of `data()`.
2. `data_unchecked()` and `data_opt()` is redundant. They can already be achieved by using `data().unwrap()` and `data().ok()`, respectively for the exact same effect. `data().ok()` is only a character longer and `data().unwrap()` is actually shorter, so brevity is not harmed here.

I propose to mark `data_unchecked()` and `data_opt()` as deprecated on the next version bump, with a message nudging the user to more idiomatic patterns.

Examples:

```rust
ctx.data::<D>() // No warnings
ctx.data_unchecked::<D>() // Use `data::<D>().unwrap()` instead.
ctx.data_opt::<D>() // Use `data::<D>().ok()` instead.
```

### DIFF
--- a/benches/static_schema.rs
+++ b/benches/static_schema.rs
@@ -187,7 +187,7 @@ impl<'a> Human<'a> {
 
     /// The friends of the human, or an empty list if they have none.
     async fn friends<'ctx>(&self, ctx: &Context<'ctx>) -> Vec<Character<'ctx>> {
-        let star_wars = ctx.data_unchecked::<StarWars>();
+        let star_wars = ctx.data::<StarWars>().unwrap();
         star_wars
             .friends(self.0)
             .into_iter()
@@ -229,7 +229,7 @@ impl<'a> Droid<'a> {
 
     /// The friends of the droid, or an empty list if they have none.
     async fn friends<'ctx>(&self, ctx: &Context<'ctx>) -> Vec<Character<'ctx>> {
-        let star_wars = ctx.data_unchecked::<StarWars>();
+        let star_wars = ctx.data::<StarWars>().unwrap();
         star_wars
             .friends(self.0)
             .into_iter()
@@ -266,7 +266,7 @@ impl QueryRoot {
         )]
         episode: Option<Episode>,
     ) -> Character<'a> {
-        let star_wars = ctx.data_unchecked::<StarWars>();
+        let star_wars = ctx.data::<StarWars>().unwrap();
         match episode {
             Some(episode) => {
                 if episode == Episode::Empire {
@@ -284,7 +284,7 @@ impl QueryRoot {
         ctx: &Context<'a>,
         #[graphql(desc = "id of the human")] id: String,
     ) -> Option<Human<'a>> {
-        ctx.data_unchecked::<StarWars>().human(&id).map(Human)
+        ctx.data::<StarWars>().unwrap().human(&id).map(Human)
     }
 
     async fn humans<'a>(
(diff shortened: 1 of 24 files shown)


xamgore:
> `*_unchecked` is a name usually reserved for unsafe functions that can potentially invoke undefined behavior

Well, it's true for the standard library.

[Dynamic enforcement with opt-out](https://rust-lang.github.io/api-guidelines/dependability.html#dynamic-enforcement-with-opt-out) chapter from the official API guideline says `async-graphql` is ok:

> The convention is to mark these opt-out functions with a suffix like `_unchecked`

patrickariel:
Thanks for the link, I think that's a good reference for the `_unchecked` convention in general.

> #### Dynamic enforcement
> Validate the input as it is processed (or ahead of time, if necessary).
> ...
> The unchecked functions can be used judiciously in cases where (1) performance dictates avoiding checks and (2) the client is otherwise confident that the inputs are valid.

But I don't really see any checks or validations being bypassed in `data_unchecked`, at least not in the sense described in the document. All it does is that it unwraps the `Option<T>` produced by the checked variant.

I would love to see if there are any counter-examples in std or other well-established libraries, though. But as far as I can tell, `_unchecked` functions are not supposed to merely be a panicking wrapper over the checked counterpart. If anything, `_unchecked` functions tend to *not* panic (because assertions are bypassed).

negezor:
This would require a lot of changes to all codebases. I'm a bit against these changes, as I myself use a similar approach with ctx.data_unchecked() as I find it more concise than the .data().unwrap() chain.

patrickariel:
To clarify, the PR only marks these methods as deprecated, it does not remove them. I'll leave it to the discretion of the maintainer as to when it actually gets removed. Existing codebases will still function just fine, only printing a harmless lint warning when compiled.

If you want to safely and accurately migrate in an automated fashion, you can use [`ast-grep`](https://github.com/ast-grep/ast-grep):

```bash
ast-grep --interactive --lang rs --pattern '$A.data_unchecked::<$T>()' --rewrite '$A.data::<$T>().unwrap()'
ast-grep --interactive --lang rs --pattern '$A.data_opt::<$T>()' --rewrite '$A.data::<$T>().ok()' 
```

This will start an interactive, AST-based search and replace throughout your codebase. These are the same commands I used for this PR (double-checked, ofc).

IMO, both `data_unchecked` and `unwrap` should be heavily discouraged in production. Personally, I'd recommend using context extensions:

```rust
pub trait MyContextExt {
    fn my_data(&self) -> Result<&MyData, MyError>;
}

impl MyContextExt for Context<'_> {
    fn my_data(&self) -> Result<&MyData, MyError /* <-- Use your app's custom error type here */> {
        self.data::<MyData>().map_err(|_| {
            MyError::new(
                "This is a properly sanitized error message containing no sensitive details.",
            )
        })
    }
}
```
Then in your query:

```rust
#[Object]
impl Query {
    async fn my_query(&self, ctx: &Context<'_>) -> Result<String, MyError> {
        let my_data: MyData = ctx.my_data()?; // <-- As concise as it gets! :)
        Ok("done".into())
    }
}
```

I can't emphasize enough how *unconventional* `async-graphql`'s `data_unchecked` is here ( [`unwrap_unchecked`](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_unchecked) exists, which makes it all the more confusing). It really jumped out at me when I first read the docs, which initially discouraged me from adopting this library. It gave me the impression that the library encourages heavy usage of unsafe functions throughout user code.

PR #1628, opened 2024-11-11
dedupe error handling

mxfactorial:
just let websocket handle it

### DIFF
--- a/integrations/axum/src/subscription.rs
+++ b/integrations/axum/src/subscription.rs
@@ -295,9 +295,7 @@ where
         futures_util::pin_mut!(stream, sink);
 
         while let Some(item) = stream.next().await {
-            if sink.send(item).await.is_err() {
-                break;
-            }
+            _ = sink.send(item).await;
         }
     }
 }
--- a/integrations/axum/src/subscription.rs
+++ b/integrations/axum/src/subscription.rs
@@ -268,13 +268,6 @@ where
             .stream
             .take_while(|res| future::ready(res.is_ok()))
             .map(Result::unwrap)
-            .filter_map(|msg| {
-                if let Message::Text(_) | Message::Binary(_) = msg {
-                    future::ready(Some(msg))
-                } else {
-                    future::ready(None)
-                }
-            })
             .map(Message::into_data);
 
         let stream =


sunli829:
When `sink.send` fails, it means the websocket has been closed, and I understand that the next call to `stream.next` will also return an error and end the loop, but I want to know the benefit of doing this. 🙂

mxfactorial:
> but I want to know the benefit of doing this

you preserve the protocol

which is helpful when
1. your client sends a [stop](https://github.com/async-graphql/async-graphql/blob/7a1391bd8539e0581fe7c0cd371a0d38718e5aea/src/http/websocket.rs#L385)
2. but your subscription resolver polls so aggressively that its blocked from reading the stop (fixed by a [10 milli sleep](https://github.com/systemaccounting/mxfactorial/blob/cbcd813d71ccd496cf16d44f162ca5c326d02cdc/services/graphql/src/main.rs#L240))
3. then the client stops accepting messages
4. and the axum integration breaks the subscription because it receives an error instead of a stop

now your debug statement in the stop branch of the websocket submodule doesnt print because of the early termination (debug lines highlighted @[1:00](https://drive.google.com/file/d/1Dnfthd7-rUgybnDMxsbvRi04mqRdhnNI/view) dont print)

so you ask yourself "wheres my stop"

sunli829:
Maybe you should use an [asynchronous WebSocket client](https://docs.rs/websocket/0.27.1/websocket/client/async/index.html), `measure_socket.read()` blocks polling and this can cause weird things to happen.

mxfactorial:
will consider that as an improvement to the sleep thank you

sunli829:
I believe that all the problems you encounter are solved when you use asynchronous WebSocket client. 🙂

mxfactorial:
this pr intends preventing the integration from violating the protocol when weird things happen

and youve received evidence of how others can benefit when the protocol remains

but this is your project and the choice to extend that benefit to others is yours

welcome to close if you prefer not to

sunli829:
I can't reproduce the problem, when the client sends `Stop`, everything works fine, you can test it with this example.

https://github.com/async-graphql/examples/tree/8f0450eec914aa546e5a36c67586b01cf8ffe087/axum/subscription

mxfactorial:
can join you in the rabbit hole of producing another blocking subscription

but easier if you just run your tests with my change and merge the reduction in error handling on [principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)

sunli829:
Does the problem still exist after switching to an asynchronous WebSocket client?

mxfactorial:
doesnt matter

please merge to remove duplicate error handling and unnecessary filter

or close if you judge your code to be an exception to principle

sunli829:
I'm closed this PR because I think the current implementation follows the protocol, feel free to open it again if you have any questions.

Issue #1498, opened 2024-03-19
Using generics with both SimpleObject and InputObject as field in output type fails


kilork:
Hi, I encountered a weird issue I want to overcome somehow. Actually, I would be happy with w/a, because as alternative I have to generate quite massive structure with a lot of fields and I do not want to do this.

## Expected Behavior

This code should work normally.

```rust
    #[tokio::test]
    pub async fn test_both_input_output_result_generic() {
        #[derive(SimpleObject, InputObject)]
        #[graphql(concrete(name = "MyObjectU32", params(u32)))]
        #[graphql(concrete(name = "MyObjectString", params(String)))]
        struct MyObject<T: InputType + OutputType> {
            a: T,
        }

        #[derive(SimpleObject)]
        struct MyObjectResult {
            fields: MyObject<u32>,
        }

        struct Query;

        #[Object]
        impl Query {
            async fn my_object_result(&self, fields: MyObject<u32>) -> MyObjectResult {
                MyObjectResult { fields }
            }
        }

        let _ = Schema::new(Query, EmptyMutation, EmptySubscription);
    }
```

## Actual Behavior

It fails with the message:

```
Register `MyObjectU32` as `Object`, but it is already registered as `InputObject`
```

## Steps to Reproduce the Problem

1. Have two structures, one generic with both SimpleObject and InputObject and second using first one, but only with SimpleObject annotation.
2. Use first structure as input parameter and second - as output in method declaration for some query.
3. Try to build a scheme with this query.

## Specifications

- Version: 7.0.3
- Platform: n/a
- Subsystem: derive macros

carlocorradini:
Any update?
I think the best solution is to allow
 `input_name` in `concrete`:
```rs
 #[graphql(concrete(..., input_name="..."))]
```
This allows us to have both generic `SimpleObject` and `InputObject` with custom names avoiding conflicts

PR #1629, opened 2024-11-12
Allow non-native concrete types in generic structs deriving SimpleObject + InputObject

dbousque:
Currently, only native types are supported in generic structs deriving SimpleObject + InputObject. So the following works:

```rust
#[graphql(concrete(name = "MyObjectString", params(String)))]
struct MyObject<T: InputType + OutputType> {
    a: T,
}
```

But this doesn't:

```rust
#[derive(Clone, Copy, PartialEq, Eq, Enum, serde::Serialize)]
enum MyEnum {
    Option1,
    Option2,
}

#[graphql(concrete(name = "MyObjectMyEnum", params(MyEnum)))]
struct MyObject<T: InputType + OutputType> {
    a: T,
}
```

This is because both input and output types have the same name, so we get a runtime error like this:
```
Register `MyObjectMyEnum` as `InputObject`, but it is already registered as `Object`
```

This is the issue reported here: https://github.com/async-graphql/async-graphql/issues/1498

This PR allows users to explicitly specify the input name of concrete types, so that the input and output types can have different names. Like so:

```rust
#[graphql(concrete(name = "MyObjectMyEnum", input_name = "MyObjectMyEnumInput", params(MyEnum)))]
struct MyObject<T: InputType + OutputType> {
    a: T,
}
```

### DIFF
--- a/derive/src/args.rs
+++ b/derive/src/args.rs
@@ -117,6 +117,8 @@ impl FromMeta for GenericParamList {
 #[derive(FromMeta)]
 pub struct ConcreteType {
     pub name: String,
+    #[darling(default)]
+    pub input_name: Option<String>,
     pub params: PathList,
     #[darling(default)]
     pub bounds: GenericParamList,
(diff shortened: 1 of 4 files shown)


sunli829:
Thanks, this is very useful. 🙂

dbousque:
Thanks @sunli829. I fixed the book example, the tests should go through now.

PR #1631, opened 2024-11-19
feat: Add `custom-error-conversion` feature

cmtm:
# Context
async-graphql doesn't support crate users implementing `From<T> for async_graphql::Error` for their own error type `T` because there's an existing impl for types that implement Display. This means that users aren't able to implement custom logic when converting their type such as setting error extensions.

The way to set error extensions is with the `ErrorExtensions`. But that requires calling `.extend()` in all the field resolvers, and prevents use of the `?` operator, since that won't call `.extend()`.

# Solution
Add `custom-error-conversion` feature which prevents `From<T: Display + ...>` from being implemented for `async_graphql::Error`. This allows users to implement their own `From<T> for Error`, which can construct Error as they need.

This has the added benefit of creating a more seamless transition path for users coming from juniper, as juniper allows custom error conversion with the `IntoFieldError` trait.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ playground = []
 raw_value = ["async-graphql-value/raw_value"]
 uuid-validator = ["uuid"]
 boxed-trait = ["async-graphql-derive/boxed-trait"]
+custom-error-conversion = []
 
 [[bench]]
 harness = false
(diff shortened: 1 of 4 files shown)


cmtm:
What are your thoughts on this @sunli829 ?

Improved naming suggestions for the feature flag are very welcome.

sunli829:
released in `v7.0.15`

PR #1632, opened 2024-11-20
use cruet to use updated regex dependency

david7a68:


david7a68:
Thought cruet used a newer version of regex. Found out immediately after opening this that I was incorrect. Closing as redundant. Apologies for the noise.

PR #1637, opened 2024-11-27
Add new feature: `nullable-result`

XiNiHa:
This PR adds a new Cargo feature, `nullable-result`.

## What does it do?

When the feature is enabled, all fields that return `Result` become nullable, and `null` is returned in place when the `Result` has `Err()`.

## Why do this?

Null bubbling is a nightmare in most cases, which makes [nullable-by-default schema one of the best practices for designing GraphQL schemas](https://graphql.org/learn/best-practices/#nullability). However, the existing behavior of `async-graphql` doesn't match with it: error bubbles until it meets `Option` to "capture" it as a null. This makes it too easy to break the app just by not wrapping it with `Option`, and wrapping all return types from resolvers that return `Result` with `Option` is a very tedious, repetitive, and easy-to-miss task to do. This wouldn't have been the case at all if `Result` got resolved into a nullable type.

If we also look at the ecosystem, Caliban, a GraphQL server framework for Scala, also infers types of fields from the return type of resolvers and emits nullable types if resolvers return a throwable type like `ZIO[R, Throwable, A]` (somewhat similar to `Result<A, Throwable>`.) [(ref)](https://ghostdogpr.github.io/caliban/docs/schema.html#schema-generation)

While I personally believe that this should be the default for all cases, I guarded it with a Cargo feature for now to keep backward compatibility. I hope we can enable this as default at some point.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ dynamic-schema = []
 graphiql = ["dep:askama"]
 raw_value = ["async-graphql-value/raw_value"]
 boxed-trait = ["async-graphql-derive/boxed-trait"]
+nullable-result = []
 
 [[bench]]
 harness = false
--- a/derive/src/complex_object.rs
+++ b/derive/src/complex_object.rs
@@ -133,7 +133,7 @@ pub fn generate(
                         .into());
                     }
                 };
-                let ty = ty.value_type();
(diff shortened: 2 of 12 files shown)


XiNiHa:
@aumetra I'd like to have this and #1638 merged. Can you take some time to review these?

PR #1640, opened 2024-12-04
Use web_time instead of std::time

omarabid:
This replaces std::time with web_time, specifically problematic for Web Socket/Subscriptions. Integrations are not replaced, since these do not support a Wasm back-end anyway.


omarabid:
@sunli829 Any chance this makes it to the next release?

Issue #1516, opened 2024-04-28
Overflow when using `MergedObject`


Sytten:
## Expected Behavior

I upgraded from v5 to v7.0.2, it should work still.

## Actual Behavior

```
overflow evaluating the requirement `http::Request<reqwest::async_impl::body::ImplStream>: std::marker::Send`
consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`proxy_api`)
required because it appears within the type `(Request<ImplStream>, Callback<Request<ImplStream>, Response<Body>>)`
required for `Chan<Envelope<Request<ImplStream>, Response<Body>>, Semaphore>` to implement `std::marker::Sync`
the full type name has been written to '/Users/sytten/Projects/Caido/proxy-backend/target/debug/deps/proxy_api-c28335827c5bb19d.long-type-8384231238600946750.txt'
required for `Arc<Chan<Envelope<Request<ImplStream>, Response<Body>>, Semaphore>>` to implement `std::marker::Send`
the full type name has been written to '/Users/sytten/Projects/Caido/proxy-backend/target/debug/deps/proxy_api-c28335827c5bb19d.long-type-17041696055801546513.txt'
required for `std::ptr::Unique<hyper::client::pool::Idle<hyper::client::client::PoolClient<reqwest::async_impl::body::ImplStream>>>` to implement `std::marker::Send`
required because it appears within the type `((Scheme, Authority), Vec<Idle<PoolClient<ImplStream>>>)`
required for `RawTable<((Scheme, Authority), Vec<Idle<PoolClient<ImplStream>>>)>` to implement `std::marker::Send`
the full type name has been written to '/Users/sytten/Projects/Caido/proxy-backend/target/debug/deps/proxy_api-c28335827c5bb19d.long-type-8221124596057829378.txt'
required for `std::sync::Mutex<hyper::client::pool::PoolInner<hyper::client::client::PoolClient<reqwest::async_impl::body::ImplStream>>>` to implement `std::marker::Sync`
1 redundant requirement hidden
required for `std::sync::Arc<std::sync::Mutex<hyper::client::pool::PoolInner<hyper::client::client::PoolClient<reqwest::async_impl::body::ImplStream>>>>` to implement `std::marker::Sync`
required for `std::sync::Arc<reqwest::async_impl::client::ClientRef>` to implement `std::marker::Send`
[...]

Issue #1506, opened 2024-03-30
Stack overflow after upgrade to 7.0.2


negezor:
I've been experiencing stack overflow in runtime, I didn't immediately notice this problem as I hadn't run tests for a long time during local development. 

After hours of comparing changes, I found out that it happened after an upgrade to 7.0.2. After spending some more time trying to figure out what broke between 7.0.1 and 7.0.2 I found the cause.

The problem turned out to be the following PR https://github.com/async-graphql/async-graphql/pull/1468. 

## Expected Behavior

Working code as on 7.0.1

## Actual Behavior
When calling GraphQL mutation and then executing a database transaction, application crashes

```
thread 'actix-rt|system:0|arbiter:7' has overflowed its stack
fatal runtime error: stack overflow
fish: Job 1, 'cargo run' terminated by signal SIGABRT (Abort)
```

Using [backtrace-on-stack-overflow](https://crates.io/crates/backtrace-on-stack-overflow) I got the following stacktrace.

<details>
<summary>Stacktrace</summary>

```
Stack Overflow:
   0: backtrace_on_stack_overflow::handle_sigsegv
   1: <unknown>
   2: <(A,B,C,D,E,F,G,H,I,J,K,L,M,N,O,P,Q,R,S,T,U) as nom::branch::Alt<Input,Output,Error>>::choice
   3: nom::branch::alt::{{closure}}
   4: <F as nom::internal::Parser<I,O,E>>::parse
   5: <(A,B,C,D,E,F,G,H,I,J,K,L,M,N,O,P,Q,R,S,T,U) as nom::branch::Alt<Input,Output,Error>>::choice
   6: nom::branch::alt::{{closure}}
   7: <F as nom::internal::Parser<I,O,E>>::parse
   8: <(A,B,C,D,E,F,G,H,I,J,K,L,M,N,O,P,Q,R,S,T,U) as nom::branch::Alt<Input,Output,Error>>::choice
   9: nom::branch::alt::{{closure}}
  10: <F as nom::internal::Parser<I,O,E>>::parse
  11: <(A,B,C,D,E,F,G,H,I,J,K,L,M,N,O,P,Q,R,S,T,U) as nom::branch::Alt<Input,Output,Error>>::choice
  12: nom::branch::alt::{{closure}}
  13: <F as nom::internal::Parser<I,O,E>>::parse
[...]

PR #1641, opened 2024-12-05
Async trait

Sytten:
Fixes https://github.com/async-graphql/async-graphql/issues/1516
Fixes https://github.com/async-graphql/async-graphql/issues/1506

Right now a lot of people me included can't upgrade to v7 because the boxed version of the trait was removed in https://github.com/async-graphql/async-graphql/pull/1468

Unfortunately the rust community didn't come up with a nice way to deal with the case where a library wants to offer both boxed and non-boxed traits. I tried in the async-trait https://github.com/dtolnay/async-trait/issues/274 and in the dead crate trait-variant (https://github.com/rust-lang/impl-trait-utils/issues/35).

I don't have the bandwidth to create and maintain such a crate, so the best I could come up that works is this.
It is mostly clean except for the trait declaration itself where you often need two implementations.


sunli829:
Although I will increase the stack size and limit the query depth to avoid stack overflow issues, `boxed-trait` seems like a good temporary solution.

Issue #1562, opened 2024-07-18
Parsed incorrect operation pos


Expurple:
## Expected Behavior

The linked reproducer should print something like following, with accurate and unique operation positions:

```text
Pos(99:6) Some(Name("create_purchases"))
Pos(136:1) Some(Name("check_services_in_period"))
Pos(49:1) Some(Name("create_sales"))
Pos(1:1) Some(Name("_setup"))
Pos(122:1) Some(Name("create_period"))
```

## Actual Behavior

Instead, it prints the following:

```text
Pos(99:6) Some(Name("create_purchases"))
Pos(136:1) Some(Name("check_services_in_period"))
Pos(49:1) Some(Name("create_sales"))
Pos(1:1) Some(Name("_setup"))
Pos(136:1) Some(Name("create_period"))
```

where the operation before last (`create_period`) has the pos from the last operation (`check_services_in_period`).

This messes with my in-house test framework which runs the operations one-by-one in order (by `pos`).

The issue seems very query-specific. I encountered it when I added `mutation create_purchases` into the middle of the existing file. If I remove it again, the issue no longer reproduces. And if I try to remove details about my endpoints from the reproducer, this also stops reproducing the issue.

## Steps to Reproduce the Problem

1. Download the reproducer: [async-graphql-repro.zip](https://github.com/user-attachments/files/16281381/async-graphql-repro.zip)
2. `cargo run` it

## Specifications

- Version: async-graphql 7.0.7
- Platform: Ubuntu 24.04 x64

Expurple:
Fixed by  #1648

PR #1648, opened 2024-12-19
Fix position calculator for unicode symbols

xamgore:
#1562 is caused by Unicode characters. In the example provided, positions were drifted. N bytes (returned from pest-parser) are not the same as N chars in the source text.

```rust
for _ in 0..pos - self.pos {    // N times
    match self.input.next() {   // read a char
        ...
    }
}

let slice = self.input[..bytes_to_read];  // N bytes
for ch in slice.chars() {                 // read a char (1-byte, 2-byte, 3-byte, 4-byte)
    match ch {
        ...
    }
}
```

### DIFF
--- a/parser/src/pos.rs
+++ b/parser/src/pos.rs
@@ -3,7 +3,6 @@ use std::{
     cmp::Ordering,
     fmt,
     hash::{Hash, Hasher},
-    str::Chars,
 };
 
 use pest::{iterators::Pair, RuleType};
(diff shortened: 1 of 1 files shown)


xamgore:
Here is an example with symbols having 2,3,4 bytes thus causing the drift.

```graphql
mutation mut {
    doSomething(
        # слова на языке отличным от English будут приводить к неправильным позициям
        form: { word: "你好" }
    ) {
        id
    }
}

query {   # should be at line 10, but .pos = 12
  entities { id }
}   
```

xamgore:
@sunli829 this bag affects the order of execution in batching operations, is it possible to merge it? What else should be done to proceed?

PR #1657, opened 2025-01-16
Update error messages for character length validation

kenkoooo:
Revise error messages to accurately reflect character count instead of byte length in validation functions.


sunli829:
thanks

PR #1663, opened 2025-01-29
Update graphql annotation directive property to support paths

kschat:
When attempting to wrap the `#[graphql]` annotation in my own proc macro, I found that it was impossible to make my macro hygienic due to the `directive` argument not accepting a full path to the directive to apply. This updates that property to use the full path to the directive type when generating code.


parser/src/parse/generated.rs
```diff
@@ -4,4 +4,4 @@
 #![allow(unused_attributes)]
```
kschat, line 4: This file changed when running the tests locally before making any changes. I'm not sure if that's expected or if there's something I need to do differently when running the tests.

PR #1667, opened 2025-02-06
integrations/axum: Add tracing support for subscription

chitoku-k:
This PR adds `tracing` feature to the `async-graphql-axum` crate. With this support, function calls from `Subscription`s are properly in the correct span.


aumetra:
Ah nice! Thanks!

PR #1669, opened 2025-02-07
fix validation failure with variables in multiple operations

cmtm:
When a document with multiple operations has variables with different types, a validation error would occur, because the validation checked that the type of the variable matched the type of *all* places it's used. In multi-operation documents, variables only apply to the selected operation.

Don't check argument types that are variables when in unselected operations.


cmtm:
Al alternative implementation would be to not do any validation on non-selected operations. That would be a larger functionality change, but would make the code simpler.

cmtm:
bump on this for comment/review

cmtm:
@aumetra, think we can get this merged in? I'm happy to implement the alternative that I suggested.

aumetra:
I'll have a look later today!

aumetra:
This change definitely makes sense! It's fine as-is, I was gonna check through the codebase for potential cleanups anyway :D
Thanks for the patch

PR #1671, opened 2025-02-12
chore(deps): update zxcvbn

negezor:


negezor:
Depends on this PR https://github.com/async-graphql/async-graphql/pull/1672, since the tracing library is now required directly and therefore is not build.

PR #1670, opened 2025-02-12
chore(deps): bump opentelemetry to 0.28.0

negezor:
The tests are passing, no other breaking changes were noticed.

PR #1672, opened 2025-02-12
chore(deps): bump non-breaking dependencies

negezor:
There are a few old versions that are duplicated in the dependency tree of the final application.

I also think it might be worthwhile to abandon support for the CBOR format in the future, since `serde_cbor` has not been maintainedfor a long time.
https://osv.dev/vulnerability/RUSTSEC-2021-0127

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,13 +60,13 @@ indexmap.workspace = true
 mime = "0.3.17"
 multer = "3.0.0"
 num-traits = "0.2.18"
-pin-project-lite = "0.2.14"
-regex = "1.10.4"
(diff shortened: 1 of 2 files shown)


uncledu:
@negezor  any updates on this pr? come from #1670

aumetra:
These minor dependency bumps can be done by each crate consumer by running `cargo update`, I do not see any actions we have to take from our side here.

As for the tracing fix and internal semver breaking updates, those are already merged and published. But thank you for the PR regardless!

And about the CBOR support, I'm planning on trimming down some features next major version.

PR #1679, opened 2025-03-01
Expanded generic section in the book into its own page

parrottq:
I just learned about `TypeName` and though I should add a section to the book.

Changes:
- Moved generics out of `SimpleObject` page since I had a hard time finding it and because it applies to more than just `SimpleObject`.
- Added a section about `TypeName` to the generic page.

Tell me if you see any typos, want something reworded, or want this section moved somewhere else.


sunli829:
thanks

PR #1683, opened 2025-03-10
replace ok_or with ok_or_else

llamington:


negezor:
Does this really improve performance? Have any measurements been taken? I think adding a callback function would cost more.

sunli829:
thanks!

PR #1685, opened 2025-03-17
Refactor `CacheStorage::get` to return `Option<V>` instead of `Option<&V>`

### LINKED ISSUE #1684: Make `CacheStorage::get` return `Option<T>` instead of `Option<&T>`
## Description of the feature

I'm attempting to integrate `tinyufo` as a `CacheStorage` implementation. However, this is currently infeasible because the `CacheStorage::get` method requires a return type of `Option<&T>`, while `tinyufo::get` returns `Option<T>`.

I tried to make `tinyufo` return `Option<&T>` but found it hard to implement. Then I looked at `async-graphql` and found there's no extra cost if changes were done on `async-graphql` side.

## Code example (if possible)

The current implementation of `CacheStorage` immediately clones the value returned by `get`:

https://github.com/async-graphql/async-graphql/blob/60e353ef7494e449ac4502c44bb37cb7f713e77c/src/dataloader/mod.rs#L334-L336

Therefore, I propose that the return type of `CacheStorage::get` be changed to `Option<T>`, as this would be functionally equivalent and allow for compatibility with `tinyufo`.

Yesterday17:
Closes #1684

### DIFF
--- a/src/dataloader/cache.rs
+++ b/src/dataloader/cache.rs
@@ -27,7 +27,7 @@ pub trait CacheStorage: Send + Sync + 'static {
 
     /// Returns a reference to the value of the key in the cache or None if it
     /// is not present in the cache.
-    fn get(&mut self, key: &Self::Key) -> Option<&Self::Value>;
+    fn get(&mut self, key: &Self::Key) -> Option<Self::Value>;
 
     /// Puts a key-value pair into the cache. If the key already exists in the
     /// cache, then it updates the key's value.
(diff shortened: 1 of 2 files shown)


avborup (commented):
I think the failing test will pass if you rebase onto `master`. v7.0.16 updated the MSRV to 1.83.

aumetra:
Looks good! Thanks! (I actually need that myself because I'm refactoring the internals to use `scc::HashCache` for a lock-free cache)

Issue #1465, opened 2024-01-15
Disable introspection, returns null in __typename--Dynamic Schema


ct-porco:
## Expected Behavior
"__typename" field is populated when introspection is disabled.
```json
{
  "data": {
    "user": {
      "avatar": {
        "__typename": "Image"
      },
      "__typename": "User"
    }
  }
}
```

## Actual Behavior
"__typename" field is always `null` when introspection is disabled, which may break queries which return `Union` types.
```json
{
  "data": {
    "user": {
      "avatar": {
        "__typename": null
      },
      "__typename": null
    }
  },
}
```

## Steps to Reproduce the Problem

1. Use dynamic schema
2. Use `schema.disable_introspection()`
3. Make a request with `__typename`

## Specifications

- Version: 6.0.3
- Platform: Debian 12 (Bookworm)
- Subsystem: N/A

hnnasit:
We are running into a similar issue as well. This works for static schema even when introspection is disabled. I've created a PR which enables a function for allowing introspection only on the `__typename` field. @sunli829 could you help review the changes in [PR](https://github.com/async-graphql/async-graphql/pull/1686) to fix the issue?

sunli829:
This is a bug, __typename should not resolve to null even with introspection disabled, fixed in `v7.0.16`.

hnnasit:
Thanks for the quick fix and doing a release @sunli829 !

PR #1686, opened 2025-03-18
feat: Allow Introspection of __typename field only for dynamic schema

hnnasit:
Issue: https://github.com/async-graphql/async-graphql/issues/1465

As described in the linked issue, `__typename` field is null when introspection is disabled for dynamic schema. This is blocking us right now.


sunli829:
This is a bug, __typename should not resolve to null even with introspection disabled, fixed in `v7.0.16`.

Issue #1545, opened 2024-06-19
Does the project support @requiresScopes for apollo


RustLoverTheCoder:
Does the project support @requiresScopes for apollo
[Apollo authorization](https://www.apollographql.com/docs/router/configuration/authorization)

PR #1695, opened 2025-04-08
feat: add @requiresScopes support

msierks-pcln:
Hello 👋

This PR adds support for the [@requiresScopes directive](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#requiresscopes).

Closes https://github.com/async-graphql/async-graphql/issues/1545.

### DIFF
--- a/derive/src/args.rs
+++ b/derive/src/args.rs
@@ -235,6 +235,8 @@ pub struct SimpleObjectField {
     #[darling(default, multiple, rename = "directive")]
     pub directives: Vec<Expr>,
     pub complexity: Option<Expr>,
+    #[darling(default, multiple)]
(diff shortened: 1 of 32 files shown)


sunli829:
thanks

Adam-Alj:
Hi, just wanted to say that this is an enterprise feature [(source)](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#requiresscopes) and likely should not be included by default in the linked directives here.

We probably want an `enable enterprise federation directives` flag or something like that to conditionally include these. I'll open an issue against this but felt it best to leave this comment on the change. @sunli829 @msierks-pcln

Issue #1703, opened 2025-05-03
GraphiQL v2 playground broken.


abdavid:
## Expected Behavior

Playground renders without any issues

## Actual Behavior

Completely white page with errors thrown in the console. 

## Steps to Reproduce the Problem

1. Enable playground
2. Start server
3. Navigate to playground

## Specifications

- Version: 7.0.16
- Platform: MacOS

## Additional details

Due to the nature of the error this indicates changes in the cdn version of GraphiQL. I seems that they quite recently pinned a newly released version. See this https://github.com/graphql/graphiql/pull/3909 that recently merged. 

My suggestion is to migrate to GraphiQL v4 EMS playground. I have done this locally myself for now to get the playground working again.

sergiimk:
Before migrating to v4 I would recommend a quick fix to pin the dependency, e.g. instead of:
```
https://unpkg.com/graphiql/graphiql.min.js
```
use
```
https://unpkg.com/graphiql@3/graphiql.min.js
```
at the very least pin to major, but even better would be to allow pinning to exact via parameter.

yasamoka:
For anyone facing the issue while waiting for the fixes to be merged, you can rely on either:

- [graphiql-version](https://github.com/yasamoka/async-graphql/tree/graphiql-version) fork if you want to run GraphiQL v3.9.0 with React v17.
- [graphiqlsource-react](https://github.com/yasamoka/async-graphql/tree/graphiqlsource-react) fork if you want to run GraphiQL v4.0.1 with React v18.

If you have dependencies that in turn depend on the published version of `async-graphql`, you can modify and run one of the following snippets then replace your use of `GraphiQLSource` in rendering a `String` by the output:

- [graphiql-version](https://github.com/yasamoka/async-graphql/tree/graphiql-version):

```
use async_graphql::http::GraphiQLSource;

#[tokio::main]
async fn main() {
    println!(
        "{}",
        GraphiQLSource::build()
            .version("3.9.0")
            .endpoint("/")
            .finish()
    );
}
```

<details>
<summary>Output</summary>
```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="robots" content="noindex">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <meta name="referrer" content="origin">

    <title>GraphiQL IDE</title>

    <style>
      body {
        height: 100%;
        margin: 0;
        width: 100%;
        overflow: hidden;
      }

[...]

PR #1704, opened 2025-05-03
Add GraphiQLSource version

yasamoka:
GraphiQL v4.0.0 was [released today](https://github.com/graphql/graphiql/commit/d1395f987b3f9c70b69ec5ad7283c63594dd7602). However, due to some issue, https://unpkg.com/graphiql currently returns "Not found: /graphiql@4.0.0". This breaks GraphiQL loading from an endpoint using `GraphiQLSource`, since `GraphiQLSource` always fetches the latest version of GraphiQL from unpkg.

This PR adds support for specifying the version of GraphiQL that `GraphiQLSource` would link into the generated HTML.

Related issue: #1703

Thank you in advance.


yasamoka:
Seems like this was an issue of errors in rendering GraphiQL v4 rather than `GraphiQLSource` being unable to fetch the latest version, since `GraphiQLSource` fetches https://unpkg.com/graphiql/graphiql.min.js, which was already working. Still, controlling the version would sidestep such issues in the future while a fix is prepared to support the latest version.

yasamoka:
A new major version of GraphiQL may require a newer version of React than the one being used in `GraphiQLSource` (#1705). This is a strong indication that we need to pin the major version of GraphiQL being used.

The user can still specify the version of GraphiQL they want to use at their own risk. This also protects against breakage of semantic versioning in the future.

I can rebase this on top of #1705 once approved.

penso:
@yasamoka might want to fix that lint issue for the author to merge faster?

We also are running into that issue.

yasamoka:
> @yasamoka might want to fix that lint issue for the author to merge faster?
> 
> We also are running into that issue.

CI is failing due to other code I did not touch.

sunli829:
thanks

PR #1707, opened 2025-05-08
Document extension context fields

xamgore:
Custom extensions may require modifications to context fields.

To avoid mistakes, it’s important to understand their connection to the request lifecycle—specifically when, where, and how each field is created, modified, and dropped.

These comments aim to clarify that lifecycle and usage context.


sunli829:
thanks

PR #1714, opened 2025-05-16
use preferred mime-type

### LINKED ISSUE #1708: Switch to application/graphql-response+json media type
[GraphQL over HTTP](https://graphql.org/learn/serving-over-http/#response-format) says that the preferred media type for response is `application/graphql-response+json` while `application/json` is legacy.  The main benefit other than not being a legacy framework is we will also have the ability to use different HTTP Status codes for various types of errors.  Obviously, this will need changes to the Error response to support specifying the response code.

zimirza:
This is a simple patch that replaces the `application/json` mime-type with the preferred `application/graphql-response+json` for graphql responses.

Closes #1708

### DIFF
--- a/integrations/actix-web/src/request.rs
+++ b/integrations/actix-web/src/request.rs
@@ -216,7 +216,7 @@ impl Responder for GraphQLResponse {
                 },
             ),
             _ => (
-                "application/json",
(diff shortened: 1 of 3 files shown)


sunli829:
thanks

PR #1727, opened 2025-06-17
Allow accessing data registered at the Schema level

ipetkov:
Sometimes it can be useful to access the global data registered in a Schema (via SchemaBuilder) to compute some data to be inserted in a Request before processing it. This data itself is not global, and should only persist for the request. Being able to access global data here avoids having to needlessly carry around the data a second time.


aumetra:
Looks good. I might actually also have a use for that in my own projects

PR #1730, opened 2025-06-24
Mark generated interface methods as #[allow(missing_docs)]

ipetkov:
Otherwise using interface derives will fail in modules/crates built with `#[deny(missing_docs)]`


aumetra:
Makes sense. Thanks!

PR #1736, opened 2025-07-16
Don't double box subscription streams

ipetkov:
Constructing a new stream internally will produce a boxed stream, except the type signatures were written as returning `impl Stream` which requires calling `.boxed()` another time (when implementing `Executor`) to satisfy the type system, but it results in two allocations (and two pointer chases during polling) at runtime unnecessarily


aumetra:
Thanks!

PR #1738, opened 2025-07-25
fix: repeatable Directive Output in SDL Generation

kay-schecker:
### Summary

This PR corrects the SDL (Schema Definition Language) output for directives that are marked as **repeatable**.

### Details

- Ensures that the keyword `repeatable` is correctly included in the directive SDL representation when `is_repeatable` is set to `true`.
- The output now conforms to the GraphQL specification for repeatable directives.

### Additions

- Added unit tests to verify correct SDL rendering:
  - ✅ `test_repeatable_directive_dsl`
  - ✅ `test_non_repeatable_directive_dsl`

### Motivation

Previously, the SDL output omitted the `repeatable` keyword even when the directive was defined as repeatable. This led to incorrect or incomplete schema definitions. With this fix, the SDL generation is now accurate and spec-compliant.

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -775,6 +775,11 @@ impl MetaDirective {
             .map(|location| location.to_value().to_string())
             .collect::<Vec<_>>()
             .join(" | ");
+
+        if self.is_repeatable {
+            write!(sdl, " repeatable").ok();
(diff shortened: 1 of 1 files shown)


aumetra:
Thanks!

PR #1739, opened 2025-07-25
Add From Option<T> impl for ConstValue

elftausend:
This allows users of the dynamic schema creation method to just use the same syntax for `Option<T>` as for `T`, which is especially handy for lists with nullable items.

Example
```rust
// ...
Ok(Some(Value::from(vec![Some(char.id), None])))
```


aumetra:
Thanks!

Issue #1531, opened 2024-06-04
Serializing custom json values wraps all numbers in `$serde_json::private::Number`


lennartkloock:
## Expected Behavior

Assume the following setup:

```rs
#[derive(Debug, serde::Deserialize, serde::Serialize)]
struct JsonMap(pub serde_json::Value);

async_graphql::scalar!(JsonMap);

#[derive(Debug, async_graphql::SimpleObject)]
struct SomeObject {
    map: JsonMap,
}
```

Serializing `SomeObject { map: JsonMap(serde_json::json!({ "n": 2 }) }` should return this json object:

```json
{
    "map": {
        "n": 2
    }
}
```

## Actual Behavior

But it actually returns this json object:

```json
{
    "map": {
        "n": {
          "$serde_json::private::Number": "2"
        }
    }
}
```

## Specifications

- Version: `7.0.5`
- Platform: Linux

sunli829:
I can't reproduce, the `SimpleObject` macro doesn't implement `Serialize` for `SomeObject`, this code doesn't compile.

xamgore:
@lennartkloock could it be `arbitrary_precision` feature of the `serde_json` crate, and not `async-graphql` itself?

frederikhors:
@lennartkloock did you find a way?

lennartkloock:
@frederikhors I honestly don't remember what I did to fix it. I'm guessing I just ended up not using this approach at all.

PR #1744, opened 2025-08-18
Introduce serialization for `u128` and `i128` via `serde_json/arbitrary_precision`.

vnermolaev:
This PR adds serializer functions to support `u128` and `i128`. The new functionality is feature-gated under `arbitrary_precision`, which also enables `serde_json/arbitrary_precision` to avoid lossy conversion when targeting JSON.

Deserializer changes are not required: the present deserializer _already_ accepts `u128` and `i128` via [the existing forwarding macro](https://github.com/async-graphql/async-graphql/blob/082201ccd040726410ac1b8a6cf6dc4a1c5179ec/value/src/deserializer.rs#L125-L129):
```rust
forward_to_deserialize_any! {
    bool i8 i16 i32 i64 i128 u8 u16 u32 u64 u128 f32 f64 char str string
    bytes byte_buf unit unit_struct seq tuple
    tuple_struct map struct identifier ignored_any
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ raw_value = ["async-graphql-value/raw_value"]
 uuid-validator = ["uuid"]
 boxed-trait = ["async-graphql-derive/boxed-trait"]
 custom-error-conversion = []
+arbitrary_precision = ["async-graphql-value/arbitrary_precision"]
 
 [[bench]]
 harness = false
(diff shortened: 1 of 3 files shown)


vnermolaev:
It turns out that enabling the `arbitrary_precision` causes some inconsistency in the `async-graphql` serialization, leading to leaking private types, e.g.,
```shell
Object(
  {
    Name(
      "$serde_json::private::Number",
    ): String(
      "255",
    ),
  },
),
```
 The leakage is exactly what issue #1531 describes.

I am closing this PR because it is incomplete.

PR #1746, opened 2025-08-20
Don't bail early on the first error when resolving containers

ipetkov:
Instead, resolve all fields fully and collect the errors in the context. This allows the request to return partial data (and multiple errors) as allowed by the GraphQL spec without needing to manually juggle this by hand.


aumetra:
Makes sense, it's spec compliant, and we can have some partial failures. Nice. Thanks! (I mean even though this *does* change the behaviour slightly, I don't foresee any semver breakage with this behaviour change so it will probably make it into a patch release)

PR #1747, opened 2025-08-22
Update the dependencies

IvanUkhov:
This is the result of running the following command:

```shell
cargo upgrade -i allow && cargo update
```

I just had to adjust the usage of `tracing` and `zxcvbn`.


aumetra:
Upgraded all the crates I can upgrade without a semver bump. Will try to cut a release first, then do the major version bumps.

Issue #1647, opened 2024-12-18
error: queries overflow the depth limit!


frederikhors:
Can you help me with this error?

```
   Compiling app v0.1.0 (C:\project\src\bin\app)
error: queries overflow the depth limit!
  |
  = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`app`)
  = note: query depth increased by 130 when computing layout of `{async block@async_graphql::resolver_utils::container::Fields<'_>::add_set<async_graphql::types::query_root::QueryRoot<graphql::Query>>::{closure#1}}`
```

SindriFr:
I also have this issue. It seems to occur for me when annotating async-graphql resolvers with #[instrument].
I think it takes a bunch of resolvers to trigger. I count 137 resolvers across QueryRoot and Mutation types in the generated schema.
The error is basically identical to what @frederikhors described.

I have an example that fails to compile with the `#[instrument]` annotation but will compile without it.
```
#[Object]
impl ChatMutation {
    #[graphql(guard = "PermissionGuard::new(Scope::Component, Operation::Read)")]
    #[instrument(skip(self, ctx, input))]
    pub async fn send_message(
        &self,
        ctx: &Context<'_>,
        input: ChatInput,
    ) -> Result<ChatResponse, Error> {
        // fn body
   }
}
```
Let me know if there's anything I can do to help!

SindriFr:
I found a workaround for this issue for anyone encountering this in the future. The compilation issue can be resolved by splitting the root object into smaller sets of merged objects, see [PR](https://github.com/async-graphql/async-graphql/pull/1757/files) for a simple example.

The root cause can be addressed by rewriting the `MergedObject` derive macro in a non-recursive manner.

frederikhors:
@SindriFr I read your last comment a while ago but didn't look into it further. Today I solved a serious problem thanks to your advice. Thank you very much!

@aumetra, I think this is an immediate first step to solve, as @SindriFr suggests:

> The root cause can be addressed by rewriting the `MergedObject` derive macro in a non-recursive manner.

aumetra:
I'll have to take a look at the impl. I'm coming back from 39C3 right now and will fix the regressions on master regarding the switch from `Inflector` first and then look at this.

dgrijalva:
Pushed a small PR to address this issue. We've recently started bumping into this. Hope it's helpful. Thanks for the great library.

PR #1757, opened 2025-09-30
doc: Document workaround for #1647

SindriFr:
Document how to resolve compilation error caused by deeply recursive calls to when using  on a type with many sub-objects.
It involves splitting up the root object into smaller merge sets, reducing the recursion depth and load on the compiler.

related issue: #1647


aumetra:
Thanks!

PR #1763, opened 2025-11-28
Implement Default, Clone, Copy for EmptyFields

j-tai:
This adds implementations of Default, Clone, and Copy to `EmptyFields`, making it consistent with `EmptyMutation` and `EmptySubscription`.


aumetra:
Thanks!

Issue #1745, opened 2025-08-20
`jiff` support


xamgore:
Many crates in the Rust ecosystem already support [jiff](https://github.com/BurntSushi/jiff), and it's time for `async-graphql`.

PRs are heavily welcomed.

xamgore:
@GodTamIt thanks!

PR #1778, opened 2026-01-07
feat: Add support for `jiff`

GodTamIt:
This adds a new optional feature `jiff` that adds scalar types for `jiff` internal types.

Fixes: #1745

### DIFF
--- a/src/types/external/jiff.rs
+++ b/src/types/external/jiff.rs
@@ -95,11 +95,12 @@ impl ScalarType for Span {
 
 #[cfg(test)]
 mod tests {
-    use crate::{ScalarType, Value};
     use jiff::{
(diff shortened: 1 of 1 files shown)


aumetra:
I think this makes sense to add before v8 and the 'great split' (where I planning to move a lot of external crates into an async-graphql-extra crate (name pending) so the core can stay stable for longer).

aumetra:
Okay, took a bit, looks alright to me. Can you please reformat it and add a CHANGELOG entry?

GodTamIt:
@aumetra thanks for fixing this up! Sorry I didn't get around to it

aumetra:
No problem at all. Thanks for implementing the support

Issue #1716, opened 2025-05-23
Dataloader freeze when dockerised through github actions


andreievg:
## Expected Behavior

Dockerised app using async-graphql app does not freeze requests when dockerised with github actions

## Actual Behavior

Requests are frozen when async-graphql app is dockerise with github actions

## Steps to Reproduce the Problem

1. Run 
```
docker run -p 9091:8000 andreievg/checking-async-graphql:with-7.0.16
``` 
2. Run 
```
curl -X POST http://127.0.0.1:9091/graphql -H 'Content-Type: application/json' -d '{ "query": "query { user(id: 1) {__typename id name } }" }'
```
3. Run again
```
curl -X POST http://127.0.0.1:9091/graphql -H 'Content-Type: application/json' -d '{ "query": "query { user(id: 1) {__typename id name } }" }'
```
4. See freeze

Docker image built from [simple example with one loader](https://github.com/andreievg/async-graphql-check/blob/4cc2c807a79de5e71f92657b9eecf1fed0cd9b27/src/main.rs#L1-L121), with [this github action](https://github.com/andreievg/async-graphql-check/blob/4cc2c807a79de5e71f92657b9eecf1fed0cd9b27/.github/workflows/dockerise.yaml#L1-L52), and [this dockerfile ](https://github.com/andreievg/async-graphql-check/blob/4cc2c807a79de5e71f92657b9eecf1fed0cd9b27/Dockerfile#L1-L28), the dockerfile has extra stuff (sorry I copy pasted to try replicate in haste)

(I am also using `--platform linux/amd64` to start docker, on m3 mac, but i was also getting freezes when running on x86-64 and amd65)

## Specifications

- Version:
- Platform:
- Subsystem: 

## Suggested solutions

I tried with my fork, replacing [futures_channel::Delay with tokio::sleep](https://github.com/andreievg/async-graphql/commit/a8df009f5c235ceb23c1208b7c69fa205a36c279) and the problem went away,[ with this change in test repo](https://github.com/andreievg/async-graphql-check/commit/42b12377393596740923662771d87d8865d1a39e), built docker image and launched with this.
```
docker run --platform linux/amd64 -p 9091:8000 andreievg/checking-async-graphql:tokio-sleep
```

I am happy to make PR, need some guidance:

1 - Replace as per [commit](https://github.com/andreievg/async-graphql/commit/a8df009f5c235ceb23c1208b7c69fa205a36c279)
[...]

PR #1780, opened 2026-01-08
fix: Use runtime timers when available

GodTamIt:
This change makes a new type internal to `async-graphql` that uses the native async runtime's timer when available. This is more efficient and can avoid spawning a new helper thread (how runtime-agnostic timers are done).

This also switches to using the `async-io` crate instead of `futures-timer`, which is better maintained. Also, if a thread is spawned, `async-io` can use it for other I/O outside of timers.

Fixes: #1716


aumetra:
Thank you! That is actually a change I wanted to do myself. I really don't like `futures-timer`

aumetra:
Needs a reformat :D

aumetra (approved):

GodTamIt:
@aumetra made some changes to the latest patchset, so just a heads up:

- We need the `rt` feature to be able to check whether we have a runtime attached (for `tokio::runtime` access). Because it no longer maps 1:1, I renamed the feature to `tokio-timer`. Lmk if you've got a better name for the new feature.
- Oops, I wasn't on nightly so not all the formatting worked. I switched to nightly here so formatting should be fixed.

aumetra:
Thanks again for the PR!

PR #1781, opened 2026-01-10
feat(tracing): add trace_scalars option to Tracing and OpenTelemetry

### LINKED ISSUE #1767: Add `trace_scalars` option to tracing extensions to prevent span explosion
## Problem

When tracing GraphQL queries, **span explosion** can occur because every field resolution creates a span. For a typical query returning an object with 20 scalar fields, this generates 20+ spans per object. With nested objects or lists, this quickly becomes thousands of spans per request.

For example, this simple query:
```graphql
query {
  users {       # 100 users
    id          # scalar
    name        # scalar  
    email       # scalar
    createdAt   # scalar
  }
}
```
Would generate **400+ field resolution spans** (100 users × 4 fields), plus parent spans.

This causes:
- Significant performance overhead
- Noise that obscures meaningful traces

## Prior Art: graphql-ruby

The Ruby GraphQL ecosystem solved this problem with a `trace_scalars` option that **defaults to `false`**.

**graphql-ruby implementation** (`lib/graphql/tracing/platform_trace.rb`):
```ruby
def initialize(trace_scalars: false, **_options)
  @trace_scalars = trace_scalars
end

def execute_field(query:, field:, **kwargs)
  return_type = field.type.unwrap
  trace_field = if return_type.kind.scalar? || return_type.kind.enum?
    (field.trace.nil? && @trace_scalars) || field.trace
  else
    true  # Always trace object types
  end
  
  if trace_field
    # Create span
  else
    super  # Skip span creation
  end
end
```

This is now the **de facto standard** in the Ruby ecosystem and is used by Datadog's official `ddtrace` gem.

## Proposal for async-graphql

Add a `trace_scalars` option to `Tracing` and `OpenTelemetry` extensions:

```rust
pub struct OpenTelemetry<T> {
    tracer: Arc<T>,
    trace_scalars: bool,  // New option, default: false
}

impl<T: Tracer> OpenTelemetry<T> {
    pub fn new(tracer: T) -> Self {
        Self {
            tracer: Arc::new(tracer),
            trace_scalars: false,  // Default to false
        }
    }
    
    pub fn with_trace_scalars(mut self, trace_scalars: bool) -> Self {
        self.trace_scalars = trace_scalars;
        self
    }
}
```

In the `resolve` method:
```rust
async fn resolve(&self, ctx: &ExtensionContext<'_>, info: ResolveInfo<'_>, next: NextResolve<'_>) -> ServerResult<Option<Value>> {
    // Check if we should trace this field
    let should_trace = if !self.trace_scalars {
        let concrete_type = MetaTypeName::concrete_typename(info.return_type);
        !ctx.schema_env
            .registry
            .types
            .get(concrete_type)
            .map(|t| t.is_leaf())  // is_leaf() returns true for Scalar/Enum
            .unwrap_or(false)
    } else {
        true
    };
    
    if should_trace && !info.is_for_introspection {
        // Create span and trace
    } else {
        next.run(ctx, info).await
    }
}
```

## Why This Should Be in Core

While #394 concluded that Extensions API is flexible enough for users to implement this themselves, I believe `trace_scalars` should be a first-class option because:

1. **It's a best practice** - graphql-ruby defaults to `trace_scalars: false` for good reason
2. **It prevents a common pitfall** - New users often don't realize tracing will explode their spans
3. **It's simple to implement** - The `ResolveInfo` and `Registry` already provide all needed type information
4. **It's backward compatible** - Can default to current behavior (`true`) or adopt the Ruby convention (`false`)

## Related

- #394 - Earlier discussion about tracing configurability
- graphql-ruby implementation: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/platform_trace.rb

I'm happy to submit a PR if this proposal is accepted.

uuushiro:
- Closes #1767

## Summary

Add `trace_scalars` option to `Tracing` and `OpenTelemetry` extensions to prevent span explosion by skipping span creation for scalar/enum field resolutions.

- When `false` (default): spans are not created for fields returning scalar or enum types
- When `true`: all field resolutions are traced (previous behavior)

This follows the pattern established by [graphql-ruby's `trace_scalars` option](https://github.com/rmosolgo/graphql-ruby/blob/215e7fa464d7e881ea2fddc81f92065181d4e029/lib/graphql/tracing/platform_trace.rb).

## API

```rust
// Existing usage (unchanged, backwards compatible)
.extension(Tracing)

// New: configure scalar tracing
.extension(Tracing::config().with_trace_scalars(true))

// OpenTelemetry
.extension(OpenTelemetry::new(tracer).with_trace_scalars(true))
```

## Changes

- `Tracing`: Kept as unit struct for backwards compatibility
- `TracingConfig`: New configurable struct returned by `Tracing::config()`
- `OpenTelemetry<T>`: Added `trace_scalars` field with `with_trace_scalars()` builder method
- Both use `MetaTypeName::concrete_typename()` and `MetaType::is_leaf()` to detect scalar/enum types

## Test plan

- [ ] Verify `Tracing` unit struct still works as before
- [ ] Verify `Tracing::config().with_trace_scalars(true)` enables scalar tracing
- [ ] Verify `OpenTelemetry::new(tracer)` defaults to no scalar tracing
- [ ] Verify `OpenTelemetry::new(tracer).with_trace_scalars(true)` enables scalar tracing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # [7.1.0] 2026-01-02
 
+- Add `trace_scalars` option to Tracing and OpenTelemetry extensions [#1781](https://github.com/async-graphql/async-graphql/pull/1781)
 - Replace async-stream in favour of asynk-strim
 - Add `From<Option<T>>` for `ConstValue` [#1739](https://github.com/async-graphql/async-graphql/pull/1739)
 - Don't bail early on first error when resolving containers [#1746](https://github.com/async-graphql/async-graphql/pull/1746)


aumetra:
Ooo I like that API design. Also I think I'm alright with changing the default to not tracing scalars in a non-breaking release since I do think that pretty much nobody should actually have depended on the scalars being traced in their Grafana dashboards (or other automations)

aumetra (changes requested):
Looks nice! Can you add a CHANGELOG entry please?

uuushiro:
@aumetra 
Thanks! I've added a CHANGELOG entry.

aumetra (approved):

Issue #186, opened 2020-06-24
Use fully qualified paths in macro expansion


Koxiaet:
Currently, this fails to compile:

```rust
mod std { }

#[async_graphql::SimpleObject]
struct Something { }
```

If the macro expansion used `::std::` instead of `std::` then it would (at least in the 2018 edition). This applies to `async_graphql::` too - writing `mod async_graphql {}` breaks it.

sunli829:
This is a hard bug to spot. 😂

phated:
@sunli829 can that be caught by enabling: https://doc.rust-lang.org/unstable-book/language-features/proc-macro-hygiene.html#proc_macro_hygiene ?

sunli829:
> @sunli829 can that be caught by enabling: https://doc.rust-lang.org/unstable-book/language-features/proc-macro-hygiene.html#proc_macro_hygiene ?

This should only support the nightly version, right?

phated:
@sunli829 we can run tests with +nightly to check in CI

sunli829:
I don't think this will solve the above problem because the stable version doesn't work. Maybe I misunderstood?

PR #1782, opened 2026-01-12
fix: macro hygiene

kschat:
## Problem Statement

I'm working on a crate that uses and re-exports `async_graphql` and expects users of this crate to _not_ use `async_graphql` directly (but rather, use the version re-exported). Due to this, when users of this crate use any of the proc macros provided by `async_grapqhl`, they run into path not found errors. The mitigation to this is to add `use my_crate::async_graphql{self, ...}`, which is a non-obvious solution since this type is required internally by the proc macros.

## Solution

Fix macro hygiene issues.

When referencing non-`async_graphql` types, the proc macros are very good with macro hygiene. However, when referencing types internal to `async_graphql`, the paths were missing the root `::`. I've updated all public proc macros to use a  `::` prefix for `async_grapqhl` types. I _believe_ I've found them all. 

Further, I've added a new property to the public proc macros, `crate`. This works exactly like the [`crate` property in `serde`](https://serde.rs/container-attrs.html#crate):

> Specify a path to the serde crate instance to use when referring to Serde APIs from generated code. This is normally only applicable when invoking re-exported Serde derives from a public macro in a different crate.

## Open Questions

I'm not quite sure what the use-case for `internal` is, but it seems to maybe solve a similar problem by changing the path for internal types from `async_graphql::` to `crate::`. This doesn't solve my problem, though. What should be done if both `internal` and `crate` are provided to the same proc macro? Currently, `internal` is checked first and `crate` is ignored if it's set as that seemed like the most backwards-compatible solution. Please let me know the preferred solution if it differs from the current.

## Issue

This was actually reported in #186, and it was _mostly_ fixed. However, the `async_graphql` paths were not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # Unreleased
 
 - fix: Use runtime timers when available [#1780](https://github.com/async-graphql/async-graphql/pull/1780)
+- Add `trace_scalars` option to Tracing and OpenTelemetry extensions [#1781](https://github.com/async-graphql/async-graphql/pull/1781)
+- Macro hygiene [#1782](https://github.com/async-graphql/async-graphql/pull/1782)
 
 # [7.1.0] 2026-01-02
 
-- Add `trace_scalars` option to Tracing and OpenTelemetry extensions [#1781](https://github.com/async-graphql/async-graphql/pull/1781)
 - Replace async-stream in favour of asynk-strim
 - Add `From<Option<T>>` for `ConstValue` [#1739](https://github.com/async-graphql/async-graphql/pull/1739)
 - Don't bail early on first error when resolving containers [#1746](https://github.com/async-graphql/async-graphql/pull/1746)


kschat:
I can't seem to be able to reproduce the build error for the "Test book examples" check locally. I'm using the same version of Rust, and mdbook. I'm running the exact same `cargo build ...` command, and `mdbook` command. It completes successfully for me locally. Any help reproducing this would be appreciated.

aumetra:
That is really weird. Also seems not related with your code changes. In general I've seen weird behaviour around serde lately with cargo. I would actually just look through that and merge it as is, then have a look into what the compiler does.

kschat:
CHANGELOG updated. Note, it appeared that [#1781](https://github.com/async-graphql/async-graphql/pull/1781/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR9) mistakenly added its change to the 7.1.0 release section, however I don't believe it was released with that version. I moved it to the "Unreleased" section.

FYI I was finally able to reproduce the test book example error. I had to remove my local `Cargo.lock` file and then the build steps caused the error. I dove into the verbose output of `cargo build ...` to see what was causing multiple versions of `serde-*.rlib` to appear. There's a divergence in the set of features used by some dependencies that's caused it to be built twice. Here are the pertinent logs:

```
     Running `/home/kyle/.rustup/toolchains/1.92.0-x86_64-unknown-linux-gnu/bin/rustc --crate-name serde_core --edition=2021 /home/kyle/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/serde_core-1.0.228/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=424 --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no -C debuginfo=2 --cfg 'feature="alloc"' --cfg 'feature="default"' --cfg 'feature="result"' --cfg 'feature="std"' --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values("alloc", "default", "rc", "result", "std", "unstable"))' -C metadata=630b93a8f1368a6d -C extra-filename=-65de9eafd919d9b6 --out-dir /home/kyle/Workspace/async-graphql/target/debug/deps -L dependency=/home/kyle/Workspace/async-graphql/target/debug/deps --cap-lints allow --check-cfg 'cfg(if_docsrs_then_no_serde_core)' --check-cfg 'cfg(no_core_cstr)' --check-cfg 'cfg(no_core_error)' --check-cfg 'cfg(no_core_net)' --check-cfg 'cfg(no_core_num_saturating)' --check-cfg 'cfg(no_diagnostic_namespace)' --check-cfg 'cfg(no_serde_derive)' --check-cfg 'cfg(no_std_atomic)' --check-cfg 'cfg(no_std_atomic64)' --check-cfg 'cfg(no_target_has_atomic)'`
```

```
     Running `/home/kyle/.rustup/toolchains/1.92.0-x86_64-unknown-linux-gnu/bin/rustc --crate-name serde_core --edition=2021 /home/kyle/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/serde_core-1.0.228/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=424 --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no --cfg 'feature="alloc"' --cfg 'feature="result"' --cfg 'feature="std"' --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values("alloc", "default", "rc", "result", "std", "unstable"))' -C metadata=603a9c8293b645a3 -C extra-filename=-9ef4b9ba2424d9a3 --out-dir /home/kyle/Workspace/async-graphql/target/debug/deps -L dependency=/home/kyle/Workspace/async-graphql/target/debug/deps --cap-lints allow --check-cfg 'cfg(if_docsrs_then_no_serde_core)' --check-cfg 'cfg(no_core_cstr)' --check-cfg 'cfg(no_core_error)' --check-cfg 'cfg(no_core_net)' --check-cfg 'cfg(no_core_num_saturating)' --check-cfg 'cfg(no_diagnostic_namespace)' --check-cfg 'cfg(no_serde_derive)' --check-cfg 'cfg(no_std_atomic)' --check-cfg 'cfg(no_std_atomic64)' --check-cfg 'cfg(no_target_has_atomic)'`
```

Diffing those logs revealed that the first entry contained `-C debuginfo=2 --cfg 'feature="default"'` while that's omitted from the second. Sorry, I reached my timebox limit and this is as far as I can go on this for now. I hope it helps.

aumetra:
Thanks for looking into it! I will do some more digging to see how I can fix it

PR #1784, opened 2026-01-19
Delete tide integration

aumetra:
`tide`, as well as the entire `async-std` project, have been dead for quite a while. This is just cleaning up the cruft.


aumetra:
Agh I need to ask for push access to the examples repo..

Issue #1773, opened 2026-01-01
Replace `Inflector` with `heck`


aumetra:
This was attempted in a patch version without semver breakage but it turned out that heck and Inflector have too many differences in how they handle word boundaries to be feasible.

This is a breaking change and will (probably) be performed next version. Every consumer will have to introspect their schemas and make sure to add renames accordingly.

PR #1785, opened 2026-01-19
Replace Inflector with heck

aumetra:
This PR replaces the deprecated and archived Inflector crate with `heck` which is at least passively maintained.

This also means that your GraphQL schema *MAY* change. Please diff and check if there are any changes and rename fields manually if you need to.


aumetra:
Closes #1773

PR #1786, opened 2026-01-19
Remove niche features

### LINKED ISSUE #1783: `password-strength-validator` feature is broken
The `min_password_strength` function has been implemented but there is no proc-macro annotation so it was never functional.

This is a candidate for removal in v8.

aumetra:
This PR does some cleanup, removing niche and/or outdated features.

- CBOR (this is niche and not used anymore)
- BSON (very MongoDB-centric. Please parse it yourself instead)
- hashbrown (this crate backs std, there is no real advantage to using it instead)
- tokio-sync (why would you deserialize a Mutex??)
- uuid-validator (this was never wired up properly. Just use the UUID type)
- zxcvbn (this was also never wired up properly. Check in application code)
- smol_str (very niche that you need small string optimization. Please newtype it)
- bigdecimal (we already have `rust_decimal` support for financial transactions. I do not see a usecase for it)

Closes #1783


aumetra:
"You wouldn't deserialize a Mutex"

Erik1000:
#619 it has been nice haha

PR #1789, opened 2026-01-21
Proper runtime abstraction

aumetra:
Is this something you could get behind @ipetkov?


ipetkov:
From a consumer perspective yes, the direction looks good! I do like that there's off the shelf implementations (e.g. `tokio`) available behind a feature flag, you have the convenience for common setups but you can also wire up your own custom thing if need be :+1:

PR #1820, opened 2026-04-01
fix: Guard on nullable field returns null instead of bubbling up

### LINKED ISSUE #798: Failed guards for nested nullable fields maybe should just return null?
Right now if a guard fails (I'm using guards for authorization) in a deep nested field, the entire query becomes null.
While having an indication of the path to the error in the "errors" field is nice, wouldn't it be even better to bubble to the nearest nullable parent like resolver errors do?
Does this makes any sense to anyone?

BTW Thanks for this library! :love_you_gesture:

### LINKED ISSUE #1532: Guard + Option<Result<T>> on a single field makes the full query null
## Expected Behavior

Using this code:

```rust
struct MyQuery;

#[Object]
impl MyQuery {
    #[graphql(name = "MyQuery")]
    async fn my_query(&self) -> Option<Result<MyQueryResult>> {
        Some(Ok(MyQueryResult { foo: "foo".to_string() }))
    }
}

#[derive(SimpleObject)]
#[graphql(complex)]
struct MyQueryResult {
    pub foo: String,
}

#[ComplexObject]
impl MyQueryResult {
    #[graphql(guard = "Gandalf")]
    async fn bar(&self) -> Option<Result<String>> {
        Some(Ok("bar".to_string()))
    }
}

struct Gandalf;

impl Guard for Gandalf {
        Err("You shall not pass".into())
    }
}
```

I expect that the following query:

```json
{
  "query": "{ MyQuery { foo, bar } }"
}
```

would return the following response:

```json
{
  "data": {
    "MyQuery": {
      "foo": "foo",
      "bar": null
    }
  },
  "errors": [
    {
      "locations": [
        {
          "column": 6,
          "line": 4
        }
      ],
      "message": "You shall not pass",
      "path": [
        "MyQuery",
        "bar"
      ]
    }
  ]
}
```

## Actual Behavior

The actual response is:

```json
{
  "data": {
    "MyQuery": null
  },
  "errors": [
    {
      "locations": [
        {
          "column": 6,
          "line": 4
        }
      ],
      "message": "You shall not pass",
      "path": [
        "MyQuery",
        "bar"
      ]
    }
  ]
}
```

## Specifications

- Version:
  ```shell
  cargo tree | grep async-graphql            
      │   │   ├── async-graphql v7.0.3
      │   │   │   ├── async-graphql-derive v7.0.3 (proc-macro)
      │   │   │   │   ├── async-graphql-parser v7.0.3
      │   │   │   │   │   ├── async-graphql-value v7.0.3
      │   │   │   ├── async-graphql-parser v7.0.3 (*)
      │   │   │   ├── async-graphql-value v7.0.3 (*)
      │   │   ├── async-graphql-axum v7.0.3
      │   │   │   ├── async-graphql v7.0.3 (*)
      ├── async-graphql v7.0.3 (*)
      ├── async-graphql-axum v7.0.3 (*)
  ```
- Platform: MacOS
- Subsystem: Intel Core i5

kenkoooo:
Fixes #798
Fixes #1532

## Problem

When a guard fails on a nullable (`Option<T>`) field, the error propagates via `?` and
causes null-bubbling to the nearest non-nullable parent, wiping out sibling field data.
Per the GraphQL spec, a field error on a nullable field should resolve that field to
`null` without affecting its parent.

## Solution

- Add an `is_option_type` helper to detect `Option<T>` return types at derive time.
- Pass a `nullable` flag into `generate_guards`.
- For nullable fields, emit `if let Err` + `ctx.add_error` + return `Value::Null`
  instead of `?`.
- Non-nullable fields retain the existing `?` propagation behaviour.

Covers `SimpleObject`, `Object`, `ComplexObject`; subscriptions always use the
non-nullable path.

## Test plan

- `test_guard_on_nullable_field_simple_object`
- `test_guard_on_nullable_field_object`
- `test_guard_on_nullable_field_complex_object`
- `test_guard_on_non_nullable_field_still_propagates`

### DIFF
--- a/derive/src/complex_object.rs
+++ b/derive/src/complex_object.rs
@@ -548,7 +548,12 @@ pub fn generate(
                 .map_err(|err| err.into_server_error(ctx.item.pos))
             };
             let async_guard = match method_args.guard.as_ref().or(object_args.guard.as_ref()) {
-                Some(code) => Some(generate_guards(&crate_name, code, async_guard_map_err, false)?),
+                Some(code) => Some(generate_guards(
+                    &crate_name,
+                    code,
+                    async_guard_map_err,
+                    false,
+                )?),
                 None => None,
             };
 
--- a/derive/src/complex_object.rs
+++ b/derive/src/complex_object.rs
@@ -14,8 +14,8 @@ use crate::{
     utils::{
         GeneratorResult, extract_input_args, gen_boxed_trait, gen_deprecation, gen_directive_calls,
         generate_default, generate_guards, get_cfg_attrs, get_crate_path, get_rustdoc,
-        get_type_path_and_name, parse_complexity_expr, parse_graphql_attrs, remove_graphql_attrs,
-        visible_fn,
+        get_type_path_and_name, is_option_type, parse_complexity_expr, parse_graphql_attrs,
+        remove_graphql_attrs, visible_fn,
     },
 };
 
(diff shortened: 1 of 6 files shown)


kenkoooo:
@aumetra Please let me know if I missed something I needed to do before pushing a PR. Thanks!

aumetra:
Oo that looks nice and fixes something I wanted to get to eventually. Thanks. I'll have a look in a bit

kenkoooo:
Hey @aumetra, hope you've been well! Just wanted to follow up on this PR since it's been a little while. Totally understand if other things have been taking priority. Just let me know if there's anything I can do to make the review easier. No pressure either way!

PR #1838, opened 2026-09-04
parser: an enum value may begin with true, false or null

bhanuc:
The specification defines [`EnumValue`](https://spec.graphql.org/October2021/#EnumValue) as *"Name but not `true` or `false` or `null`"* — the three names, not the three prefixes. The grammar excludes any name that merely **starts with** one:

```pest
enum_value = ${ !(boolean | null) ~ name }
```

`!(boolean | null)` is a lookahead for the literal alone, so `boolean` matches the `true` inside `trueish` and the name is rejected.

### Reproduction

```rust
use async_graphql_parser::parse_query;

// fails: "unexpected null"
parse_query(r#"{ f(status: nullable) }"#).unwrap();
```

Anything called `nullable`, `trueish`, `falsey` or `null_` is currently not lexable as an enum value, and a document naming one fails to parse.

This is reachable from ordinary schemas rather than being a corner case. A PostgreSQL constraint named `nullPrefixTestTable_pkey` is a legal identifier, and a GraphQL layer over such a database exposes constraint names as enum members — so an upsert naming one cannot be expressed.

### The fix

It follows the idiom already used for numbers a few lines above, which have the same "a literal must not run into a name" problem:

```pest
number = @{ (float | int) ~ !name_start }
```

Reject the literal only where it is *not* followed by a name character — that is, only where it is the whole name:

```pest
name_continue = @{ (ASCII_ALPHA | ASCII_DIGIT | "_") }
name = @{ name_start ~ name_continue* }

enum_value = ${ !((boolean | null) ~ !name_continue) ~ name }
```

`name_continue` is factored out of `name` so the two cannot drift apart.

### Tests

Added `test_enum_value_may_begin_with_a_literal` beside the existing `test_number_lookahead_restrictions`, which covers the analogous case:

| input | before | after |
|---|---|---|
| `true`, `false`, `null` | rejected | rejected |
| `nullable`, `trueish`, `falsey`, `null_`, `nullPrefixTestTable_pkey` | **rejected** | accepted |

`cargo test -p async-graphql-parser` passes, including `generated_code_is_fresh`. `generated.rs` was regenerated by `tests/codegen.rs` as that test directs, rather than edited.


bhanuc:
Filed #1839 with the background: the reproduction, why constraint and enum names from a database hit this in practice, and the two projects affected. This PR is the fix for it.

